PR #51, opened 2024-01-17
Add a `postarchive` command to fix the release artifact

dkotter:
### All Submissions:

---

When we run the command `npm run archive`, it successfully builds a zip archive of the extension but this archive does not contain the inner base folder. This prevents this archive from being used as a release asset.

Looking at some of the other Woo extensions, they have a custom `postarchive` npm command that will unzip this artifact and re-zip it with the proper base folder structure. This PR adds that same command.

1. On `trunk`, run `npm run archive`
2. Run the `zipinfo` command on the archive file that this steps create. You shouldn't see `woocommerce-square/` in the file path from that command output
3. Checkout this branch and follow the same steps as above
4. You should now see `woocommerce-square/` in the file paths

### Changelog entry

> Fix - ensure our archive command works properly.


Sidsector9 (approved):
Thanks for the PR, tests well 👍

PR #125, opened 2024-05-21
Update the Apple Developer domain verification file

### LINKED ISSUE #122: Apple Pay domain verification fails due to unexpected well-known file size
### Describe the bug

When installing WooCommerce Square on a fresh site, Apple Pay domain verification will fail. When inspecting the logs, the failure is a result of the `.well-known/apple-developer-merchantid-domain-association` file being 24 bytes larger than expected:

```

2024-05-17T23:20:20+00:00 Notice Error: Unable to verify domain freely-ruling-cardinal.jurassic.ninja - response = Array
(
    [errors] => Array
        (
            [0] => Array
                (
                    [category] => INVALID_REQUEST_ERROR
                    [code] => BAD_REQUEST
                    [detail] => Invalid domain `example.com`: expected GET https://example.com/.well-known/apple-developer-merchantid-domain-association to return 9098 bytes but instead returned 9122.
                    [field] => domain_name
                )

        )

)
```

This same error is not present on sites that have successfully verified their domain previously, even when deleting the `well-known` file and regenerating it by disabling and re-enabling digital wallets.

### To reproduce

1. Spin up a fresh site.
2. Install WooCommerce Square.
3. Connect to a Square account.
4. Assign the location.
5. Enable Square as a payment method.
6. Ensure that digital wallets are enabled.
7. Check the notice:

```
Apple Pay is not available with Square - there was a problem with registering your store domain with Square/Apple Pay. View the Square logs to find out what caused the registration to fail.
```

8. Review the logs.
9. Determine that domain verification failed because the `well-known` file is `9122` bytes.

### Expected behavior

Either:
1. The domain verification file is reduced to the expected `9098` bytes.
2. The domain verification process works when the file is `9122` bytes (like it did previously).

### Additional details

8206303-zen
8191881-zen

Given that...
- Square 4.6.1 was released over a month ago.
- Domain verification is handled outside of the plugin.
- Sites that have previously been registered do not encounter any errors with a file that is 9122 bytes.
- Other payment providers have different size files (e.g., Stripe's is 9094 bytes).

...I'd say that this is an issue with Square/Block and not Apple or our integration.

We checked with the Square team in p1715992183595699-slack-CMKCR9H1C to see if this is a configuration change they made on their end.

dkotter:
### All Submissions:

---

As reported in #122, new sites are running into issues with getting their domain verified with Apple. It seems the domain verification file we bundle with this extension needs to be updated so this PR handles that.

Note that for any sites that are already verified, it doesn't seem like this impacts them, as things still work. But new sites can't be verified without this change.

Also note that Square is recommending that we don't store this file for longer than 24 hours as it can change in the future. Not sure how we go about changing that but something we may want to consider going forward.

Just to capture how things work right now:

1. When the Square plugin is active and we are in the admin on either the Square settings page or Square Gateway settings page, we fire off a method to check the Apple Pay domain registration (`Digital_Wallet\apple_pay_domain_registration`)
2. This method checks if the gateway is enabled and if digital wallets are enabled, as well as if Apple Pay is turned on. If all match, it then runs a method to check for the existence of the verification file (`Digital_Wallet\check_apple_pay_verification_file`)
3. This method will check if the domain verification file exists on the server and if it doesn't, it will copy the version from the plugin to the server root within the `.well-known` directory. It also checks if the contents of the existing file match what is in the plugin and if not, it will copy over the new file (important for us here, as this means the new file in this PR will be copied over even for sites that already have this file in place)
4. If we have a domain verification file in place and we aren't already registered, we then run a method to register the domain with Apple (`Digital_Wallet\register_apple_pay_domain`)
5. This method makes a remote request to `https://connect.squareup(sandbox).com/v2/apple-pay/domains` to verify the domain. If something goes wrong, we log an error message but otherwise we consider the domain verified

Closes #122 

1. Create a new, publicly accessible environment and install and configure WooCommerce
2. Checkout this PR and build the Square extension (`npm install && npm run build`)
3. Install and configure the Square extension. Turn on the Square payment gateway and ensure Digital Wallets are turned on. Ensure your domain is verified on the Square side (see https://woocommerce.com/document/woocommerce-square/payment-settings/#section-6)
4. Ensure no error messages are shown around domain verification
5. Using Safari, ensure the Apple Pay buttons show as expected. Note there's an existing bug where Apple Pay doesn't show in block-based carts or checkouts so will need to test with classic integrations 

### Changelog entry

> Fix - Update the Apple Pay domain verification file.


peterwilsoncc (approved):
Updated verification file looks good to me.

Testing notes:

1. Fresh Environment, WP, WooCommerce, Square (this branch) install 
3. Set up ngrok to allow global connections
4. Activate Plugin
5. Configure Square for sandbox account
6. Activate gateway including digital wallets
7. Verify file is copied over*
8. Open Square dashboard, go to apple pay
9. Click verify domain
10. Domain verified
11. Create $1 simple, virtual product
12. Verify button appears and allows payments on product page
13. Verify button appears and allows payments on cart (shortcode) page
14. Verify button appears and allows payments on checkout (shortcode) page

\* My local environment runs with WP in a sub-directory and the file was copied to the root directory of the WordPress install rather than the site's root directory. Once I relocated the file it worked as expected. However, I think this is a bug with the way [Chassis](https://github.com/Chassis/Chassis) is set up rather than the Square extension.

PR #235, opened 2024-11-13
Multiple Variations Support

### LINKED ISSUE #6: [Enhancement] To synchronize Option name when available (instead of creating an attribute with "Attribute" name)
It's possible to create Item Options in Square with an Attribute name:

https://d.pr/i/CjqlmF 
Full Size: https://d.pr/i/CjqlmF

Square doc about that: https://squareup.com/help/us/en/article/6689-item-options

However, when importing this variable product, its Attribute is still called "Attribute":

https://d.pr/i/rrWNg0 
Full Size: https://d.pr/i/rrWNg0

These are Square logs for two variations:

```
{
            "type": "ITEM_VARIATION",
            "id": "W7TNLTQQHQ7L6ATJYPXET54W",
            "updated_at": "2020-10-08T10:18:18.966Z",
            "version": 1602152298966,
            "is_deleted": false,
            "present_at_all_locations": true,
            "item_variation_data": {
                "item_id": "WOIBYEJ4AK3VXZXM7Z3O43IO",
                "name": "Orange",
                "sku": "SKU_orange",
                "ordinal": 0,
                "pricing_type": "FIXED_PRICING",
                "price_money": {
                    "amount": 1000,
                    "currency": "USD"
                },
                "location_overrides": [
                    {
                        "location_id": "MF0A9X5YSXZAA",
                        "track_inventory": true
                    }
                ],
                "item_option_values": [
                    {
                        "item_option_id": "RSDEMJN3SMRNTSWQU4ZMTT2O",
                        "item_option_value_id": "ZOFMJVQ7EQ32MEIBII4VNRZH"
                    }
                ]
            }
        },
        {
            "type": "ITEM_VARIATION",
            "id": "KHKSK7APJJYUO5P6NOOPHEHP",
            "updated_at": "2020-10-08T10:18:18.966Z",
            "version": 1602152298966,
            "is_deleted": false,
            "present_at_all_locations": true,
            "item_variation_data": {
                "item_id": "WOIBYEJ4AK3VXZXM7Z3O43IO",
                "name": "Vanilla",
                "sku": "SKU_vanilla",
                "ordinal": 1,
                "pricing_type": "FIXED_PRICING",
                "price_money": {
                    "amount": 1500,
                    "currency": "USD"
                },
                "location_overrides": [
                    {
                        "location_id": "MF0A9X5YSXZAA",
                        "track_inventory": true
                    }
                ],
                "item_option_values": [
                    {
                        "item_option_id": "RSDEMJN3SMRNTSWQU4ZMTT2O",
                        "item_option_value_id": "TQUW55Z7FNJJEU7LQGMAOL6D"
                    }
                ]
```

I'm not sure if it's a recent Square feature or not (personally I don't remember being able to create Item Options with attribute name a year ago but I can be mistaking). 

If it's a new feature, I guess it's Enhancement. If not, it's a bug so added both tags and let you decide :)

### LINKED ISSUE #31: Product Variation title doesn't update in WC when pulled from Square
### Describe the bug

In this PR https://github.com/woocommerce/woocommerce-square-private/pull/1122 , issue related to the product title has been fixed. If a product is a variation product then the title is not getting updated of variations on sync.

Related Discussion: 
- https://github.com/woocommerce/woocommerce-square-private/pull/1122#issuecomment-1737467247
- https://github.com/woocommerce/woocommerce-square-private/pull/1122#issuecomment-1738910362 

| Square Product | WP Store after the sync |
|--------|--------|
| Screenshot 2023-09-27 at 7 32 04 PM | Screenshot 2023-09-27 at 7 31 26 PM |

### To reproduce

1. Add Product with **variations** in Square
2. As an admin, go to the site's Square settings `/wp-admin/admin.php?page=wc-settings&tab=square&section` > Sync settings > select Square > Save
3.  Click on "Import all products from Square"
4. After the import finishes, change the name of some items in Square. For example, add a string after their current name. Like changing "Car" to "Car - 1". Also update title of variations products.
5. Go to the site's Square Update page `/wp-admin/admin.php?page=wc-settings&tab=square&section=update` > click on Sync now
6. After the sync finishes, go to WooCommerce > Products `/wp-admin/edit.php?post_type=product`. 
7. Notice that the variations of product titles weren't updated.

### LINKED ISSUE #231: Sync fails when Dynamic Options used by Variations
When Woo is set as SOR, the sync fails when the dynamic options are used by a variable product. The plugin tries to update the variation name but it is not allowed when the variations use dynamic options. Even at the Square dashboard, we can not edit it.

Image

Here is the logged error:

Image

```
(
            [category:Square\Models\Error:private] => INVALID_REQUEST_ERROR
            [code:Square\Models\Error:private] => INVALID_VALUE
            [detail:Square\Models\Error:private] => Invalid object: Invalid Object with Id: XRYU7OMNCTNVD652UI4JENFD
[merchant_token=ML8ES32FTDWFZ] Item variation names may not be edited if they use item options (requested name "trunk-dynamic-options-variable-product woo02 - Red", actual name "Red"). Variation ID `XRYU7OMNCTNVD652UI4JENFD`.
            [field:Square\Models\Error:private] => 
)
```

We should skip updating the variation name here:

https://github.com/woocommerce/woocommerce-square/blob/d47dd1d162d8a9d0f6efe0b279db019556e0bba4/includes/Handlers/Product/Woo_SOR.php#L228

faisal-alvi:
### All Submissions:

---

This PR adds support for **Multiple Variations**. 

### Square and WooCommerce Attribute Types
- **Square** options:
  - **Dynamic Options**: Shared across multiple items, similar to WooCommerce’s taxonomy attributes; variation names cannot be changed.
  - **Static Options**: Unique to individual items; variation names can be changed.

- **WooCommerce** attributes:
  - **Taxonomy**: Defined globally across products.
  - **Custom**: Defined on a per-product basis.

### Key Differences Between Dynamic and Static Options
Dynamic options can be shared across products and mirror WooCommerce’s taxonomy attributes, and variation names cannot be modified, as they are automatically generated by combining the attribute values. For example: “Red, Small.” Static options are unique to each item. If multiple options exist at Square, they must be dynamic.

### Syncing Logic (Overview)
1. **WooCommerce to Square Sync**:
   - Creates **dynamic options** if there are multiple WooCommerce attributes, regardless of the WooCommerce's attributes type.
   - For a single attribute, it creates a **static option** at Square.

2. **Square to WooCommerce Sync/Import**:
   - Creates taxonomy attributes if the option slug matches an existing WooCommerce taxonomy attribute.
   - If only one option is received, it checks for an existing WooCommerce taxonomy attribute or creates a custom attribute if no match is found.

### Syncing Logic for WooCommerce and Square (Detail)

**From WooCommerce to Square:**
- If WooCommerce has more than one attribute and is set as the Source of Record (SOR), this PR will create dynamic options at Square, regardless of whether the attributes are taxonomy or custom. This is because we cannot create more than one static options for the single item (i.e product) at Square.
- If there’s only one attribute, it creates a static option at Square.
- If an attribute requires syncing, the PR creates an option at Square. New attribute values are also created at Square when synced from WooCommerce.
  
**From Square to WooCommerce:**
- If the option name received from Square matches the slug of an existing WooCommerce taxonomy attribute, this PR creates a taxonomy attribute. If the merchant updates the slug in WooCommerce, it may not match next time and a custom attribute will be created instead.
- If only one option is received, this PR checks against existing WooCommerce taxonomy attributes and attaches it if matched; otherwise, it creates a custom attribute at the product level. When creating a custom attribute, the PR reuses an existing custom attribute name, as Square does not provide the static option name. If no matching name exists, "Attribute" is assigned as the custom name.

### Square API Limitations and Solutions
Square’s API only provides the option ID used in variations, not the display name. However, to match WooCommerce taxonomy attributes or to create a custom attribute, we need the option's display name. To address this, we fetch all options via another Square API ([discussion link](https://github.com/woocommerce/woocommerce-square/issues/6#issuecomment-1881295200)) and store the option names and their IDs in a transient (`wc_square_options_data`) with a 1-day expiration. The `fetch_options_data()` function initiates this before any import or sync process.

If a transient doesn’t have updated data, a sync may fail when creating an option with a name that already exists at Square, which would require a latest options data. This PR then stops the sync and sets the `woocommerce_square_refresh_sync_cycle` option to refresh the data.

### Avoiding Attribute Name Conflicts
If WooCommerce’s taxonomy attribute is used, this PR passes the taxonomy name (e.g., “pa_color” rather than “Color”) to Square, preventing conflicts with custom attributes of the same name. The “pa_” prefix helps WooCommerce users distinguish between taxonomy and custom attributes, avoiding potential conflicts.

### New Option for Tracking
This PR introduces a new `_dynamic_options` option with `true/false` values to track whether dynamic options are used for a product. This is not in use but may be useful in future.

### Defining Variations with “Any” attributes

Noting that there's a [Defining Variations with “Any” attributes](https://woocommerce.com/document/variable-product/#variations-with-undefined-attributes) docs page for Variable Products on Woo.com that provides additional context on how this PR attempts to handle the update between WooCommerce and Square.

### Notes for Merchants
- Avoid leaving any variation dropdowns set to “Any” in products with multiple attributes, as Square does not support empty values for dynamic options. <- this one is also handled in this PR so that plugin creates "Any" option item at Square and maintains to set "Any" (means no selection) at WooCommerce and vice versa. It was decided on slack [here](https://fueled10up.slack.com/archives/C05REAQUK28/p1734108751498199?thread_ts=1733309599.907139&cid=C05REAQUK28).
- Avoid changing taxonomy attribute slugs or attribute names in WooCommerce, or renaming dynamic options in Square.
- Avoid creating custom attributes with names identical to existing taxonomy attributes, it results in issues during import/sync. Products using the same-named custom attribute are updated (replaced) incorrectly with the taxonomy attributes.
- Do not create custom attributes with names identical to existing taxonomy attributes.

### Post-Tasks

- [#251](https://github.com/woocommerce/woocommerce-square/issues/251) - Display Notices for Sync Limitations.
- [#252](https://github.com/woocommerce/woocommerce-square/issues/252) - Importing/Syncing from Square to Woocommerce, all the dynamic option values are added to the attribute, for example, for colour attribute, if there is only red and green options are selected at Square, after the import/sync, it adds all of the dynamic option values to the Woocommerce attribute regardless of what selection. This is not creating any issue to the purchases of the side, but this should be addressed.
- [#254](https://github.com/woocommerce/woocommerce-square/issues/254) - If merchant changes the name of the attribute, this will create a new option/attribute instead of using the previous one. Think of a way to handle this.

Closes #6, Closes #31 and Closes #231.

1. Set WooCommerce as SOR.
2. Create the following product types in WooCommerce:
   - Simple Product
   - Variable Product with:
     - 1 Custom Attribute
     - 1 Taxonomy Attribute
     - 2 Taxonomy Attributes
     - 2 Custom Attributes
     - 1 Custom + 1 Taxonomy Attribute
     - 2 Taxonomy Attributes + 1 Custom Attribute
3. Confirm these products are **auto-synced** and correctly created in Square.
4. Confirm the same for **manual sync** using the “Sync” button in settings.
5. Import all products from Square to WooCommerce and verify they retain attributes without issues (e.g., taxonomy attribute not converted to custom).
6. Make changes to the random products and confirm it syncs to the Square and Imports back to the Woo without and issues, few examples:
    1. add one more custom attribute to the “2 Taxonomy Attributes” product 
    2. remove 1 taxonomy attribute from the “2 Taxonomy Attributes + 1 Custom Attribute” product 
    3. add one more variation to any product 
    4. remove one variation from any product 
    5. change the dropdown of any variation in “1 Custom Attribute” product
    6. change the dropdown of all variations in “2 Taxonomy Attributes + 1 Custom Attribute” product
    7. add one more attribute value and create update the dropdown values of each variation
    8. etc.
7. Set Square as SOR and perform all of the above steps (sorry :)) This time the chagnes from the point 6 should be made at Square.
8. **Important**: Confirm no issues affect existing users by creating all product types in trunk and performing sync/import in the fix branch.

### Changelog entry

> Add - Support for multiple Product Variations sync.


includes/Handlers/Product/Woo_SOR.php
```diff
@@ -81,12 +81,75 @@ public static function update_catalog_item( CatalogObject $catalog_object, \WC_P
...
+				$options_ids            = array();
+				$options_data_transient = get_transient( 'wc_square_options_data' );
+
+				// Set the product as a dynamic options product.
+				update_post_meta( $product->get_id(), '_dynamic_options', true );
+
+				// Loop through the attributes to create options and values at Square.
+				foreach ( $attributes as $attribute_id => $attribute ) {
+
+					$attribute_name = $attribute->get_name(); //wow it gives pa_size which saved me!
```
dkotter: Assuming we don't need this comment:

```suggestion
					$attribute_name = $attribute->get_name();
```

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -225,7 +290,102 @@ public static function update_catalog_variation( CatalogObject $catalog_object,
...
+					// Check if it's a global attribute (taxonomy-based, e.g., "pa_color")
+					$taxonomy_exists = false;
+					if ( taxonomy_exists( $attribute_id ) ) {
+						// Use wc_attribute_label for global attributes
+						$attribute_name   = $attribute_id;
+						$variation_name[] = $attribute_value = get_term_by( 'slug', $attribute_value, $attribute_id )->name;
+						$taxonomy_exists  = true;
+					} else {
+						// For custom attributes, simply use the cleaned-up attribute ID
+						$attribute_name   = ucwords( str_replace( '-', ' ', $attribute_id ) ); // what if merchant changes the name?
```
dkotter: Maybe change this to a `TODO` comment to make it easier to find in the future? Or remove all together if we have this tracked elsewhere

includes/Sync/Product_Import.php
```diff
@@ -874,6 +1066,13 @@ protected function save_product_meta( $product_id, $data ) {
 			wp_set_object_terms( $product_id, $term_ids, 'product_cat' );
 		}
 
+		// Update custom meta.
+		if ( $data['custom_meta'] ) {
+			foreach ( $data['custom_meta'] as $meta_key => $meta_value ) {
+				update_post_meta( $product_id, $meta_key, $meta_value );
```
dkotter, line 1114: Do we need any sanitization on these values before updating the meta? Or is that done upstream?

dkotter (changes requested):

includes/Handlers/Product/Woo_SOR.php
faisal-alvi: Removed it from here as this would be handled in https://github.com/woocommerce/woocommerce-square/issues/254, thanks!

includes/Sync/Product_Import.php
faisal-alvi: This seems to be sanitized in core via `sanitize_meta()`.

iamdharmesh:
Hi @faisal-alvi, 

I have added initial E2E tests for the Woo SOR with products created using the following combinations:
>- Variable Product with:
>     - 1 Custom Attribute
>     - 1 Taxonomy Attribute
>     - 2 Taxonomy Attributes
>     - 2 Custom Attributes
>     - 1 Custom + 1 Taxonomy Attribute
>     - 2 Taxonomy Attributes + 1 Custom Attribute

Additionally, I have started writing E2E tests for the Square SOR. However, while working on it, I noticed an issue: if an attribute name begins with `pa_` and a global attribute with the same name does not exist, the attribute value does not get selected in the created variations. Please refer to the screenshots below for clarification:

| Square Side | WC Side |
| - | - |
| image | image |

**Steps to Reproduce:**
1. Delete all product attributes.
1. Create item options in Square with the names `pa_color` and `pa_size`, and add option values to these attributes.
1. Create a product on the Square using the created item options.
1. Import the product and check the variations in the imported product.

_If you don't want to run these steps you can simply run the test `[Square SOR] Import multiple variations products from Square` from `d8.square-sor-multiple-variations.spec.js` file and can see the product variations of the product._

Please let me know if you need any further information on this.

Thank you.

faisal-alvi:
Hi @iamdharmesh, Thank you for the detailed comment and the E2E tests—much appreciated! Regarding the issue you mentioned, the presence of the `pa_` prefix in Square typically means the option was synced from WooCommerce. This scenario can occur if the merchant deletes the corresponding taxonomy in WooCommerce after syncing.  

I've addressed this. Now, in such cases, a plugin will automatically create a new taxonomy attribute in WooCommerce if one with the `pa_` prefix doesn't exist and assign it accordingly. Please re-run the tests on your end and let me know if they work. Also, feel free to share if you notice anything else while working on the E2E tests. Once E2E tests are finished, please move the ticket further on the board. Thanks!

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -81,12 +81,75 @@ public static function update_catalog_item( CatalogObject $catalog_object, \WC_P
...
+			/**
+			 * If there are multiple variations, it must be a considered as Dynamic Options supported product.
+			 * Create/Update and Assign Dynamic Options only if a product
+			 * has multiple attributes OR options already exists in Square.
+			 */
+			if (
+				count( $attributes ) > 1
+			) {
+				$options_ids            = array();
+				$options_data_transient = get_transient( 'wc_square_options_data' );
```
iamdharmesh: I think we should have a fallback mechanism to generate the requested data in case it is not available in the transient storage. If generating the data is too heavy, we may need to consider saving the data in a more reliable location. This is because there is no guarantee that transient data will persist until its expiration time, it might disappear just seconds after being set. More information [here](https://developer.wordpress.org/apis/transients/#overview:~:text=Everyone%20seems%20to%20misunderstand%20how%20transient%20expiration%20works%2C%20so%20the%20long%20and%20short%20of%20it%20is%3A%20transient%20expiration%20times%20are%20a%20maximum%20time.%20There%20is%20no%20minimum%20age.%20Transients%20might%20disappear%20one%20second%20after%20you%20set%20them%2C%20or%2024%20hours%2C%20but%20they%20will%20never%20be%20around%20after%20the%20expiration%20time.)

iamdharmesh:
@faisal-alvi The E2E test was failing due to the idempotency key being reused for the upsert item option. I have updated the logic of key generation [here](https://github.com/woocommerce/woocommerce-square/pull/235/commits/5c1644a76f853942e325f2773fcd9b277edd29fc) and its working fine now. Please feel free to look and update if needed. Thanks.

faisal-alvi:
@dkotter over to you for code review.

dkotter (approved):

vikrampm1:
@qasumitbagthariya bumping this up in your queue to try and wrap it up asap.

qasumitbagthariya:
### QA Update  ✅ 


-------------

I have verified this PR in the `multiple-variations-support` branch which has been fixed and is functioning as intended.

Verified all the cases related to variations and the steps provided in the PR description

**Square SOR**

https://github.com/user-attachments/assets/c2013784-28c3-40f6-a2a5-253ea8de349d

**Woocommerce SOR**

https://github.com/user-attachments/assets/cfb1c89e-1e17-4a7c-9efe-82212d38fae1



**Testing Environment** 

<Details>

- WordPress: 6.7.1
- WooCommerce - 9.5.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: multiple-variations-support

</Details>


Testing Document status:

- [x] Yes

qasumitbagthariya (approved):

includes/API.php
```diff
@@ -597,6 +598,159 @@ public function retrieve_inventory_physical_count( $physical_count_id ) {
...
+		$response = $this->list_catalog( $cursor, array( 'ITEM_OPTION' ) );
+
+		if ( ! $response->get_data() instanceof ListCatalogResponse ) {
+			throw new \Exception( 'API response data is invalid' );
+		}
+
+		$objects = $response->get_data()->getObjects() ? $response->get_data()->getObjects() : array();
+
+		foreach ( $objects as $object ) {
+			$options_data[ $object->getId() ]['name'] = $object->getItemOptionData()->getDisplayName();
```
wjrosa, line 630: Can any of the getters return `null` values? If so, it would be nice to add some checks to avoid errors.

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -81,12 +81,76 @@ public static function update_catalog_item( CatalogObject $catalog_object, \WC_P
 			$item_data->setReportingCategory( $square_category );
 		}
 
-		$catalog_variations = $item_data->getVariations() ?: array();
+		$attributes = $product->get_attributes();
+
+		$product_variation_ids = $product->get_children();
+		$catalog_variations    = $item_data->getVariations() ? $item_data->getVariations() : array();
```
wjrosa: Since PHP 7.0 we can change this to:
```suggestion
		$catalog_variations    = $item_data->getVariations() ?? array();
```

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -81,12 +81,76 @@ public static function update_catalog_item( CatalogObject $catalog_object, \WC_P
...
+			 * If there are multiple variations, it must be a considered as Dynamic Options supported product.
+			 * Create/Update and Assign Dynamic Options only if a product
+			 * has multiple attributes OR options already exists in Square.
+			 */
+			if (
+				count( $attributes ) > 1
+			) {
+				$options_ids  = array();
+				$result       = wc_square()->get_api()->retrieve_options_data();
+				$options_data = isset( $result[1] ) ? $result[1] : array();
```
wjrosa: Same as above:
```suggestion
				$options_data = $result[1] ?? array();
```

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -225,7 +291,111 @@ public static function update_catalog_variation( CatalogObject $catalog_object,
...
 		if ( 'variation' === $product->get_type() ) {
-			$variation_data->setName( $product->get_name() );
+			$result                = wc_square()->get_api()->retrieve_options_data();
+			$options_data          = isset( $result[1] ) ? $result[1] : array();
+			$parent_product        = wc_get_product( $product->get_parent_id() );
+			$attributes            = $parent_product->get_attributes();
+			$variation_items       = $product->get_attributes();
+			$variation_item_values = array();
+
+			if ( 1 === count( $attributes ) ) {
```
wjrosa, line 301: I see that these changes contain multiple "actions" that could be new methods to make reading easier. Basically, every time we have a comment here, the following code could be a new method named after the explanation. But this is optional

includes/Sync/Manual_Synchronization.php
```diff
@@ -1698,6 +1743,25 @@ protected function assign_next_steps() {
...
+	 *
+	 * @throws \Exception
+	 */
+	protected function fetch_options_data() {
+		$cursor     = $this->get_attr( 'fetch_options_data_cursor' ) ? $this->get_attr( 'fetch_options_data_cursor' ) : '';
+		$result     = wc_square()->get_api()->retrieve_options_data( $cursor );
+		$new_cursor = isset( $result[2] ) ? $result[2] : null;
+
+		if ( ! empty( $new_cursor ) ) {
+			$this->set_attr( 'fetch_options_data_cursor', $new_cursor );
```
wjrosa: I believe this line can be moved outside the condition, as `$new_cursor` will be null per the default value above anyway.

includes/Sync/Product_Import.php
```diff
@@ -1056,6 +1297,8 @@ protected function save_variations( $product_id, $data ) {
 				$updated_attribute_keys = array();
 
 				foreach ( $variation['attributes'] as $attribute_key => $attribute ) {
+					// Set empty if attribute is to 'Any' to prevent it from being saved.
+					$attribute['option'] = isset( $attribute['option'] ) && 'Any' !== $attribute['option'] ? $attribute['option'] : '';
```
wjrosa: Maybe we should move `Any` to a new constant.

wjrosa (approved):
Looks good! Some multiple optional suggestions 👍

vikrampm1:
Thanks @wjrosa. @faisal-alvi see if you can make those optional suggested changes above and then we can move this to the next step.

jeffpaul:
Noting that there's a [Defining Variations with “Any” attributes](https://woocommerce.com/document/variable-product/#variations-with-undefined-attributes) docs page for Variable Products on Woo.com that provides additional context on how this PR attempts to handle the update between WooCommerce and Square.

includes/API.php
faisal-alvi: At this point in the code, there is a `getDisplayName`, which appears to be present for every item.

wjrosa (approved):
Thanks for addressing the suggestions! Looks good to me 👍

jeffpaul:
@vikrampm1 @faisal-alvi documenting some items here from our conversation with Woo & Square on 29 JAN:
* [ ] Square to craft a video tutorial for Multiple Variations to be used in merchant education / marketing
* [ ] Chris/Kevin @ Woo will look to help with merchant outreach for feedback
* [ ] Square & Woo account management & customer support to check on merchant feedback
* [x] 10up to ping Woo HEs for pre-existing merchant request feedback (done in https://github.com/woocommerce/woocommerce-square/issues/6#issuecomment-2622202560)

jeffpaul:
@faisal-alvi (and others watching this PR) note that we're awaiting confirmation to release these changes as-is and will advance this for Regression Testing once that's confirmed.

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.7.1
- Theme: Storefront 4.6.0
- WooCommerce - 9.7.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>


Next Step- Ready to Merge 🚀

PR #238, opened 2024-11-15
[E2E] Group tests by tagging them and run each group in parallel within the GitHub Actions workflow.

iamdharmesh:
### All Submissions:

---

PR organizes the E2E tests into groups using [tags](https://playwright.dev/docs/test-annotations#tag-tests). This allows each group of tests to execute in parallel within the GitHub Actions workflow, enabling faster execution and reducing the test result time.

Additionally, this PR adds a concurrency check to cancel running tests when a newer commit becomes available on the PR, ensuring that only the latest E2E tests are executed.

Verify that each test group of E2E tests runs in parallel within the GitHub Actions workflow and that all tests pass successfully.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -66,8 +66,13 @@ jobs:
 
       - name: Run E2E tests
         id: square_e2e_tests
-        if: ${{ github.event_name == 'pull_request' }}
+        if: ${{ github.event_name == 'pull_request' && matrix.type != '@general' }}
         run: npm run test:e2e -- --grep ${{ matrix.type }}
+
+      - name: Run E2E tests (General)
+        id: square_e2e_tests_general
+        if: ${{ github.event_name == 'push' && matrix.type == '@general' }}
+        run: npm run test:e2e -- --grep-invert "@cashapp|@sync|@giftcard"
 
       - name: Remove existing labels
         uses: actions/github-script@v6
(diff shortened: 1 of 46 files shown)


.github/workflows/e2e.yml
```diff
@@ -59,7 +67,7 @@ jobs:
       - name: Run E2E tests
         id: square_e2e_tests
         if: ${{ github.event_name == 'pull_request' }}
-        run: npm run test:e2e
+        run: npm run test:e2e -- --grep ${{ matrix.type }}
```
dkotter, line 70: Just one thought. If I'm following correctly here, this will only run tagged tests, correct? So if someone adds a new test and doesn't tag it (either on accident or because they don't realize it needs tagged) that test would never end up running.

Wondering about two different approaches:

1. Is there an easy way to add a test lint action that verifies a newly added test has a tag associated with it? This would ensure all tests are tagged properly before getting merged in
2. Or is there an easy way to run all non-tagged tests here?

If neither are easily possible, I'm fine proceeding as-is. For now it's fine and don't think we'll have an issue remembering to add that tag but just thinking for the future when others are working here, they may not realize they need to add that

iamdharmesh: Thanks for pointing this out, @dkotter. I have updated the test run command for the `@general` tag to use `--grep-invert`. So, now it will cover all general and non-tagged tests under the `@general` group.

dkotter (approved):

PR #313, opened 2025-03-18
Update all actions we use in our workflows to pull from specific pinned commits

dkotter:
### All Submissions:

---

In order to help protect against compromised actions, instead of including actions based on their major version (like v4), this PR switches all the actions we use to pull based on the commit hash from the latest release.

While this does impact maintenance a bit going forward, it ensures that we we're always using actions that we (hopefully) trust and if an action gets compromised (which [happens](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/)) we don't have to worry that we're using a compromised action. This also goes along with what [GitHub suggests](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

Ensure all our workflows still run as expected

### Changelog entry

> Dev - Update all third-party actions our workflows rely on to use versions based on specific commit hashes.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,4 +1,0 @@
-changelog:
-  exclude:
-    labels:
-      - "changelog: none"
--- a/.github/workflows/backlog-automation-issues.yml
+++ b/.github/workflows/backlog-automation-issues.yml
@@ -11,7 +11,7 @@ jobs:
     name: Add issue to project
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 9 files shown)


.github/workflows/release.yml
dkotter, line 1: Not sure what this workflow was used for in the past but doesn't actually contain a valid workflow so it throws an error anytime code is pushed, so figured we might as well delete it

iamdharmesh (approved):
🚀

PR #329, opened 2025-04-09
Fix/318 Restore `update_from_square()` for Price Override Support

### LINKED ISSUE #318: Review Product Price Update logic
9525647-zen

Please feel free to close this if I misunderstood the way Square sync works - but sounds like when Square is the SOR and if you don't have location override on prices - the price updates do not occur when changed on Square:

https://github.com/woocommerce/woocommerce-square/blob/45d78a2545ea26b09d7e00b10c5a394a47df1275/includes/Handlers/Product.php#L1633

1. Set Square as SOR
2. Create product on Square and sync it to Woo
3. On Square ensure you have all locations selected while creating an item
4. Change price on Square
5. Sync to Woo, see the price unchanged 
6. If you add a price override for the location set in Woo, it works

cc @iamdharmesh

faisal-alvi:
### All Submissions:

---

This PR reintroduces the use of `update_from_square()` function, which was previously removed in the Multiple Variation Support PR as its responsibilities were assumed to be handled by `$product_import->update_product()`. However, it was discovered that the price override functionality (by location) was not fully covered in `update_product()`. By restoring `update_from_square()`, this PR ensures that price overrides from Square are correctly applied during sync.

**Note for reviewers:**
The variation name has already been set in the `update_product()` ([here](https://github.com/woocommerce/woocommerce-square/blob/3f232cee954edd96abafc6f2be9570fc1405b2f1/includes/Sync/Product_Import.php#L1193)) so we are removing the name resetting from `update_from_square()`.

Closes #318.

1. Set Square as SOR
2. Create product on Square and sync it to Woo via Import All action
3. On Square ensure you have all locations selected while creating an item
4. Change price on Square
5. Sync to Woo, see the price should be updated
6. Edit product on Square add -> Use override option to add a different price -> Sync and see it updated on Woo
7. Remove the override, change the base price, it should work again.
8. Remove override and save, see it should still work.
9. Hence, try using the override with different locations and confirm it works in all scenarios.

### Changelog entry

> Fix - Issue with Product Price override update logic.

### DIFF
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -155,24 +155,6 @@ class Product {
 
 					$variation->update_meta_data( self::SQUARE_VARIATION_ID_META_KEY, $catalog_variation->getId() );
 
-					/**
-					 * Allow overriding variation name during product import from Square
-					 *
-					 * @since 3.3.0
-					 *
-					 * @param string                             $variation_name Variation name to update.
-					 * @param \SquareConnect\Model\CatalogObject $catalog_variation Catalog item variation being imported.
-					 * @param \SquareConnect\Model\CatalogItem   $catalog_item Catalog item being imported.
-					 * @param \WC_Product_Variation              $variation Variation being updated.
-					 * @return false|string String to override the variation name, false to disable updating
-					 *                      and keep existing name.
-					 * @since 3.3.0
-					 */
-					$variation_name = apply_filters( 'wc_square_update_product_set_variation_name', $catalog_variation->getItemVariationData()->getName(), $catalog_variation, $catalog_item, $variation );
-					if ( false !== $variation_name ) {
-						$variation->set_name( $variation_name );
-					}
-
 					self::update_price_money( $variation, $catalog_variation );
 
 					if ( $with_inventory && wc_square()->get_settings_handler()->is_inventory_sync_enabled() ) {
@@ -1627,7 +1609,7 @@ class Product {
 	 * @param \WC_Product|\WC_Product_Variation $product
 	 * @param \Square\Models\CatalogObject $catalog_variation
 	 */
-	private static function update_price_money( $product, \Square\Models\CatalogObject $catalog_variation ) {
+	public static function update_price_money( $product, \Square\Models\CatalogObject $catalog_variation ) {
 		$location_overrides = $catalog_variation->getItemVariationData()->getLocationOverrides();
 
 		if ( is_null( $location_overrides ) ) {
--- a/includes/Sync/Interval_Polling.php
+++ b/includes/Sync/Interval_Polling.php
@@ -223,6 +223,7 @@ class Interval_Polling extends Stepped_Job {
 
 							// Update the product, this will update/create the variations as well.
 							$product_import->update_product( $product, $data );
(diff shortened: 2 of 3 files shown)


dkotter (approved):

qasumitbagthariya:
### QA Update  ✅ 

-------------

I have verified this PR in the `fix/318` branch, which has been fixed and is functioning as intended.

- Price Override with different location

https://github.com/user-attachments/assets/1d7ae99d-0cec-42ea-8e37-13f151ad286e



**Testing Environment** 

<Details>

- WordPress: 6.7.2
- WooCommerce - 9.8.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: fix/318

</Details>


Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

kaushikasomaiya:
Hi @faisal-alvi Just a quick question -  I think the only place where price specifically gets updated is update_price_money() which has the check below

```
		if ( is_null( $location_overrides ) ) {
			return;
		}
```

I noted $location_overrides is null for newly created products 
```
'locationOverrides' => 
          array (
          )
```
while if you set it once and remove, it passes it as NULL prices which then causes the fall back price to be used
```
'locationOverrides' => 
          array (
            'value' => 
            array (
              0 => 
              \Square\Models\ItemVariationLocationOverrides::__set_state(array(
                 'locationId' => 
                array (
                  'value' => 'LQ0FQ2T0J078F',
                ),
                 'priceMoney' => NULL,
                 'pricingType' => NULL,

```

Will the fix take care of this part? As in use the base price when location overrides is not present at all initially
```
'locationOverrides' => 
          array (
          )
```

faisal-alvi:
Hey @kaushikasomaiya — good question! From what I’ve seen, the scenario you’re describing shouldn’t be an issue. I think the fix should handle both cases — whether `locationOverrides` is an empty array or contains overrides with null values — so it should still correctly fall back to the base price.

QA has already verified it across these variations, but definitely appreciate a second pair of eyes. Would be great if you could give it a spin too and let me know if you spot anything off. 🙌

kaushikasomaiya:
@faisal-alvi I am noticing there was use of `update_from_square()` in version 4.8.7 which took the price update logic to a different path - it calls update_price_money() and that will always fail updating the price near 

```
		if ( is_null( $location_overrides ) ) {
			return;
		}
```

However with the latest version which is already released 4.9.0 there is no use of `update_from_square()` and `$product_import->update_product( $product, $data );` correctly sets the price of the product. 

I am sure the current PR still fixes something related to the price override but my original concern was the case where newly created product in Square doesn't have $location_overrides set at all causing price update to be skipped however this is now fixed in 4.9.0

My guess is 4.9.0 broke the override feature but this PR will help.

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.8
- WooCommerce - 9.8.3
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>


Next Step- Ready to Merge 🚀

PR #342, opened 2025-04-17
Add Support for EMV 3-D Secure to Comply with Japan METI Regulation

### LINKED ISSUE #300: Add Support for EMV 3-D Secure
## Description

To comply with new regulatory requirements and improve payment security, this issue proposes adding EMV 3-D Secure (3DS) support. A recent update from Square Developer confirms that Japan's Ministry of Economy, Trade, and Industry (METI) has mandated that all eCommerce credit card transactions processed in Japan must comply with EMV 3-D Secure by March 31, 2025.  See more here: [https://developer.squareup.com/forums/t/emv-3-d-secure-mandate-in-japan/19294](https://developer.squareup.com/forums/t/emv-3-d-secure-mandate-in-japan/19294).

Square has introduced a new payment flow to handle buyer authentication automatically, replacing the existing verifyBuyer method. Updating the extension to support EMV 3DS will ensure compliance with Square's latest security requirements while improving fraud prevention.

## Why This Is Needed

* Regulatory Compliance – Required for processing credit card transactions in Japan from March 31, 2025.
* Improved Fraud Prevention – Enhances security for online transactions by adding an extra authentication step for buyers.
* Future-Proofing – Square will retire verifyBuyer, making this migration necessary for long-term compatibility.
* Simplified Payment Handling – Square's new payment flow automates 3DS handling, reducing the need for custom logic in the extension.

### 1\. Update Payment Flow to Support EMV 3DS

The extension should adopt Square’s new buyer verification flow, which involves:

* Calling `tokenize()` to obtain a payment token.
  * This method will now support additional fields:
    * `total`, `intent`, `billingContact`, `customerInitiated`, and `sellerKeyedin`.
  * Square will determine regulatory requirements dynamically and apply 3DS authentication as needed.
* Calling `createPayment` using the returned payment token to finalize the transaction.

### 2\. Remove Dependency on verifyBuyer

* Square’s new authentication flow eliminates the need for verifyBuyer, which will eventually be deprecated.
* The extension should remove or disable verifyBuyer-related logic to prevent conflicts.

### 3\. Test & Validate the Integration

* Ensure payments automatically trigger 3DS verification when required.
* Validate the process across:
  * One-time payments
  * Subscription-based payments (exempt cases should still be processed correctly)
  * WooCommerce checkout flow compatibility

### 4\. Provide Admin Configuration Options

* Add a setting in the extension to enable/disable EMV 3-D Secure enforcement based on merchant needs.

## Reference Documentation

* Square Developer Docs (EMV 3DS): [https://developer.squareup.com/docs/web-payments/take-card-payment](https://developer.squareup.com/docs/web-payments/take-card-payment)
* Japan 3DS Mandate (METI Announcement): [https://www.j-credit.or.jp/security/document/index.html#EMV3DS](https://www.j-credit.or.jp/security/document/index.html#EMV3DS)
* Square Developer Forums & Support: [https://developer.squareup.com/forums/](https://developer.squareup.com/forums/)

## Next Steps & Considerations

- [X] Confirm Square’s latest 3DS API requirements and validate implementation details.
- [X] Update the extension's payment flow to use `tokenize()` and `createPayment()`.
- [X] Remove outdated verifyBuyer logic and replace it with Square’s automated buyer verification.
- [X] Run testing across payment scenarios to confirm correct authentication behavior.
- [X] Ensure compatibility with WooCommerce Subscriptions and other extensions.

faisal-alvi:
### All Submissions:

---

This PR introduces support for Square’s new EMV 3-D Secure (3DS) authentication flow, in line with regulatory requirements set by Japan’s Ministry of Economy, Trade, and Industry (METI), effective March 31, 2025.

Key Updates:
- Implements Square’s new buyer verification flow using `tokenize()`.
- Removes the deprecated `verifyBuyer()` method, as Square is phasing it out.

Closes #300.
Closes https://linear.app/a8c/issue/SQUARE-35/add-support-for-emv-3-d-secure.

1. Connect your Square sandbox account.
2. Attempt a one-time payment through the WooCommerce checkout with test card details that trigger 3DS verification.
3. Test tokenization with sandbox and live cards.
4. Confirm that the order completes successfully.
5. Repeat the process for a subscription-based product to ensure compatibility with WooCommerce Subscriptions.
6. Ensure compatibility with other extensions like Pre-Orders along with the WooCommerce Subscriptions. (as mentioned [here](https://github.com/woocommerce/woocommerce-square/pull/342#pullrequestreview-2802910858))
7. Verify payment completion in both WooCommerce and Square dashboards.
8. Please move to the trunk branch and save a couple of credit cards, and then move back to the fix branch and check if the payment works with those saved credit cards without any issues.

### Changelog entry

> Add - Support for Square’s EMV 3-D Secure authentication flow to comply with Japan’s 3DS mandate and improve payment security.

### DIFF
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -132,33 +132,6 @@ export const usePaymentForm = (
 	);
 
 	/**
-	 * Generates a verification buyer token
-	 *
-	 * @param {Object} payments     Instance of Square.payments().
-	 * @param {string} paymentToken Payment Token to verify
-	 *
-	 * @return {Promise} Returns promise which will be resolved in handleVerifyBuyerResponse callback
-	 */
-	const verifyBuyer = useCallback(
-		async ( payments, paymentToken ) => {
-			let verificationResponse;
-			try {
-				verificationResponse = await payments.verifyBuyer(
-					paymentToken,
-					verificationDetails
-				);
-
-				return handleVerifyBuyerResponse( verificationResponse );
-			} catch ( error ) {
-				handleErrors( [ error ] );
-			}
-
-			return false;
-		},
-		[ verificationDetails, handleVerifyBuyerResponse ]
-	);
-
-	/**
 	 * When customers interact with the Square Payments iframe elements,
 	 * determine whether the cardBrandChanged event has occurred and set card type.
 	 *
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -53,6 +53,8 @@ export const usePaymentForm = (
 				],
 			},
 			intent,
+			customerInitiated: true,
+			sellerKeyedIn: false,
 		};
 
 		if ( intent === 'CHARGE' ) {
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -7,8 +7,7 @@ import { useState, useCallback, useMemo } from '@wordpress/element';
  * Internal dependencies
  */
 import {
-	getSquareServerData,
-	log,
+	getSquareServerData, log,
 } from '../square-utils';
 import { PAYMENT_METHOD_NAME } from './constants';
 
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -8,9 +8,7 @@ import { useState, useCallback, useMemo } from '@wordpress/element';
  */
 import {
 	getSquareServerData,
-	handleErrors,
 	log,
-	logData,
 } from '../square-utils';
 import { PAYMENT_METHOD_NAME } from './constants';
 
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -173,7 +173,6 @@ export const usePaymentForm = (
 		getPostalCode,
 		cardType,
 		createNonce,
-		verifyBuyer,
 		getPaymentMethodData,
 	};
 };
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -123,7 +123,7 @@ export const usePaymentForm = (
 	const createNonce = useCallback(
 		async ( card ) => {
 			if ( ! token ) {
-				return await card.tokenize();
+				return await card.tokenize( verificationDetails );
 			}
 
 			return token;
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -132,35 +132,6 @@ export const usePaymentForm = (
 	);
 
 	/**
-	 * Handles the response from Payments.verifyBuyer() and resolves promise
-	 *
-	 * @param {Object} verificationResult Verify buyer result from Square
-	 */
-	const handleVerifyBuyerResponse = useCallback( ( verificationResult ) => {
-		const response = {
-			notices: [],
-			logs: [],
-		};
-
-		// no errors, but also no verification token.
-		if ( ! verificationResult || ! verificationResult.token ) {
-			logData(
-				'Verification token is missing from the Square response',
-				response
-			);
-			log(
-				'Verification token is missing from the Square response',
-				'error'
-			);
-			handleErrors( [], response );
-		} else {
-			response.verificationToken = verificationResult.token;
-		}
-
-		return response;
-	}, [] );
-
-	/**
 	 * Generates a verification buyer token
 	 *
 	 * @param {Object} payments     Instance of Square.payments().
--- a/src/blocks/credit-card/use-payment-form.js
+++ b/src/blocks/credit-card/use-payment-form.js
@@ -6,9 +6,7 @@ import { useState, useCallback, useMemo } from '@wordpress/element';
 /**
  * Internal dependencies
  */
-import {
-	getSquareServerData, log,
-} from '../square-utils';
+import { getSquareServerData, log } from '../square-utils';
 import { PAYMENT_METHOD_NAME } from './constants';
 
 /**
--- a/includes/Gateway/API/Requests/Card.php
+++ b/includes/Gateway/API/Requests/Card.php
@@ -65,7 +65,7 @@ class Card extends \WooCommerce\Square\API\Request {
 		);
 
 		// 3DS / SCA verification token (from JS)
-		if ( ! empty( $order->payment->verification_token ) ) {
+		if ( ! empty( $order->payment->verification_token ) && 'saved_card' !== $order->payment->verification_token ) {
 			$request->setVerificationToken( $order->payment->verification_token );
 		}
 
--- a/includes/Gateway/API/Requests/Payments.php
+++ b/includes/Gateway/API/Requests/Payments.php
@@ -124,7 +124,7 @@ class Payments extends \WooCommerce\Square\API\Request {
 			$this->square_request->setSourceId( ! empty( $order->payment->token ) ? $order->payment->token : $order->payment->nonce->credit_card );
 
 			// 3DS / SCA verification token (from JS)
-			if ( ! empty( $order->payment->verification_token ) ) {
+			if ( ! empty( $order->payment->verification_token ) && 'saved_card' !== $order->payment->verification_token ) {
 				$this->square_request->setVerificationToken( $order->payment->verification_token );
 			}
 		}
--- a/includes/Gateway/API/Requests/Transactions.php
+++ b/includes/Gateway/API/Requests/Transactions.php
@@ -111,7 +111,7 @@ class Transactions extends \WooCommerce\Square\API\Request {
 		}
 
 		// 3DS / SCA verification token (from JS)
-		if ( ! empty( $order->payment->verification_token ) ) {
+		if ( ! empty( $order->payment->verification_token ) && 'saved_card' !== $order->payment->verification_token ) {
 			$this->square_request->setVerificationToken( $order->payment->verification_token );
 		}
 
--- a/src/blocks/credit-card/use-payment-processing.js
+++ b/src/blocks/credit-card/use-payment-processing.js
@@ -79,19 +79,7 @@ export const usePaymentProcessing = (
 			const paymentToken = paymentData.token || paymentData.nonce;
 
 			if ( paymentToken ) {
-				const verifyBuyerResponse = await verifyBuyer(
-					square.current.payments,
-					paymentToken
-				);
-
-				paymentData.verificationToken =
-					verifyBuyerResponse.verificationToken || '';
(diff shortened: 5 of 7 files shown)


dkotter:
Doing a quick review, code looks okay to me. @Sidsector9 as you have more experience here working on Square and payment handling, do you mind giving this a review to ensure the approach @faisal-alvi is taking here makes sense?

In addition @faisal-alvi, this needs to be something a user can turn on/off, with the default being off I believe (meaning we'll need a new setting for this). Is that something that is already being taken care of here?

jeffpaul:
> In addition @faisal-alvi, this needs to be something a user can turn on/off, with the default being off I believe (meaning we'll need a new setting for this). Is that something that is already being taken care of here?

Expanding on this a bit, per input from Square we'll want this enabled for any territories where it's required but for merchants in other territories where its not required (e.g US, Canada, Australia) those merchants should have the option to enable the feature in the Square settings.  Those merchants will still need to go into Square to set up Risk Manager and rules, but we can add some copy on the Square settings to note that (and perhaps deep link).

faisal-alvi:
Thanks for the review @dkotter!

Would like to highlight a few points for clarity and input:

- **Admin Setting for 3DS**: The requirement mentions an admin setting to enable/disable EMV 3-D Secure. However, based on Square’s new payment flow, this may not be possible. When `verificationDetails` are passed into `tokenize()`, Square enforces 3DS automatically, so it doesn’t appear to be something we can toggle manually. Also the `verifyBuyer()` method is going to be depcreated by Square so having an option to keep it does not worth it.
- **Global Impact**: While this mandate originates in Japan (effective March 2025), Square’s updated buyer verification approach seems to be applied globally. So this change will apply to all regions, not just Japan.
- **Credit Cards Only**: EMV 3-D Secure applies only to credit cards. For digital wallets (Apple Pay, Google Pay, etc.), we’ll continue to handle them as we currently do, without 3DS enforcement.
- **Saved Tokens**: This PR still needs some work around saved tokens. For example, the `verification_token` seems to still be present when saving the card on file, which is causing issues. I’ll need to adjust that; perhaps by omitting it entirely from the request. Would appreciate thoughts on whether that’s the right approach.
- **Pre-existing Saved Cards**: Tokens saved before this change will likely continue to work, but they may not be 3DS compliant under Square’s new requirements. Wondering if we should guide merchants to delete and re-save those cards, or if there’s a better approach here?

Would really appreciate a code review from @Sidsector9 and any feedback on the above points. Also open to suggestions for next steps or corrections if any of the above assumptions are off.

dkotter:
> The requirement mentions an admin setting to enable/disable EMV 3-D Secure. However, based on Square’s new payment flow, this may not be possible. When verificationDetails are passed into tokenize(), Square enforces 3DS automatically, so it doesn’t appear to be something we can toggle manually

We chatted with Square earlier this week and this is something they specifically requested, a setting to turn this on. If 3DS happens automatically in this new approach and there isn't a way to toggle that off, probably worth a question back to Square to ensure that sounds correct to them / cc @jeffpaul 

> Tokens saved before this change will likely continue to work, but they may not be 3DS compliant under Square’s new requirements. Wondering if we should guide merchants to delete and re-save those cards, or if there’s a better approach here?

Not sure the best approach for this, particularly for stores that have lots of recurring payments (as I imagine having them manually delete those all will be a pain). Maybe just something that is called out in the changelog or if wanting to be more aggressive, could be an admin notice.

faisal-alvi:
Thanks @dkotter. Based on what I’ve explored, technically we could make 3DS optional by keeping the existing `verifyBuyer()` flow and only calling it conditionally, that would satisfy the “toggle” requirement.

However, since Square will deprecate `verifyBuyer()`, keeping that path isn’t sustainable. Once deprecated, that fallback won’t work, which leaves us with no way to disable 3DS per merchant, the new flow enforces it automatically when `verificationDetails` are passed to `tokenize()`.

So, unless Square provides an alternate method to bypass or conditionally apply 3DS in this new flow, we may have no choice but to enforce 3DS globally for all merchants. Would it be worth checking back with Square to confirm if they have any workaround or planned flexibility here? cc @jeffpaul

jeffpaul:
@faisal-alvi the 3DS optional bit is not for folks (e.g. those in Japan) to be able to disable it but for folks in countries where 3DS is not mandated (e.g. US, Canada, Australia) to be able to enable it for their storefront.

Sidsector9:
> technically we could make 3DS optional by keeping the existing verifyBuyer() flow and only calling it conditionally, that would satisfy the “toggle” requirement.

3DS is cardholder bank (issuer)-driven. When `verifyBuyer()` is called, for frictionless cards, the 3DS challenge is not shown, while for others it is shown. In some countries, such as EU countries and UK, this is enforced by law (PSD2), and even with Risk Manager, you cannot disable it.

I am not familiar with EVM 3-D, I'll dive into this tomorrow and review the feature along with the code.

Related: https://github.com/woocommerce/woocommerce-square-private/pull/892

src/blocks/credit-card/use-payment-form.js
```diff
@@ -123,70 +123,14 @@ export const usePaymentForm = (
...
-
-	/**
-	 * Generates a verification buyer token
-	 *
-	 * @param {Object} payments     Instance of Square.payments().
-	 * @param {string} paymentToken Payment Token to verify
-	 *
-	 * @return {Promise} Returns promise which will be resolved in handleVerifyBuyerResponse callback
-	 */
-	const verifyBuyer = useCallback(
```
Sidsector9, line 171: You'll need to remove `verifyBuyer` from the object returned by `usePaymentForm()`

Sidsector9 (changes requested):
Requested few changes.

Once the PR is complete:
- Test tokenization with sandbox and live cards.
- Test with Subscriptions and Pre-Orders.

Sidsector9:
So, reading up on Square, I came up with 2 questions. Is Risk Manager an opt-in feature where:
1. 3DS can be enabled in countries where SCA is not mandatory?
2. 3DS can be disabled in countries where SCA is mandatory?

From what I understand (all speculation), in principle – Square's Risk Manager is a 3DS enablement tool, not a 3DS disablement tool.

For countries where SCA is not mandatory, merchants can opt-in to enable 3DS to improve fraud protection by setting up rules in Risk Manager. For example:
- Verify through challenge when transaction amount > $100.
- Verify through challenge when payment velocity > 5.

For countries where SCA is mandatory (e.g., the UK and EU member states), or where SCA is required by the card issuer, 3DS becomes a regulatory or issuer-enforced obligation that Square applies automatically. In such cases, Risk Manager cannot be used to disable or bypass 3DS. Risk Manager only adds additional 3DS enforcement when it's not already mandatory; it cannot remove or override authentication that is required by law or the payment network rules.

@jeffpaul @dkotter can you confirm the above with the Square team? Thanks!

Up until this PR, we were calling `verifyBuyer()` unconditionally for all transactions. In the US, where SCA is not mandatory, `verifyBuyer()` is still called, but if the card issuer does not support 3DS or determines authentication isn't needed, Square silently proceeds without a challenge. In countries where SCA is mandated by law or enforced by the issuer, `verifyBuyer()` triggers either a challenge (OTP, SMS, etc.) or performs a frictionless authentication in compliance with regulations.

In this PR, we are replacing `verifyBuyer()` with a new flow, which also unconditionally attempts SCA for all transactions. Nice work @faisal-alvi!

The only way I found to opt-in to 3DS is via the Square Risk Manager dashboard. I am not sure if this can be done through the plugin's settings screen, as I don't see an API to opt-in and set rules. Something we need to clarify with the Square team.

> Pre-existing Saved Cards: Tokens saved before this change will likely continue to work, but they may not be 3DS compliant under Square’s new requirements. Wondering if we should guide merchants to delete and re-save those cards, or if there’s a better approach here?

I don't see a reason why this wouldn't work. To confirm, I'd suggest co-ordinating with @qasumitbagthariya to tokenize a card on the `trunk` branch, and then attempt a transaction using the tokenized card on the `enhancement/300` branch.

dkotter:
> can you confirm the above with the Square team? Thanks!

Reached out to them and this does sound accurate, copying the response from them:

> Yes this is correct. We recommend invoking the 3DS flows for all payments.
>
> If the seller has not interacted with Square's Risk Manager and set up custom rules, then in non-mandated markets the inclusion of the 3DS features essentially does nothing. Payments will still go through and we won't use the additional information being passed to us to request that the issuers do the step-up authentication.
>
> If the seller has interacted with Risk Manager to set up rules, since the integration already supports collecting and passing the information needed to make 3DS work, then they get to use these features with no further updates or changes from the WooCommerce side

So I think that means we do not need a new setting on the WordPress side to turn this on or off

dkotter (approved):
This looks good to me from a code perspective. @faisal-alvi can you ensure the couple comments from Sid around testing are captured in the PR description so those scenarios are properly tested by QA?

faisal-alvi:
@dkotter Sorry about the last-minute update, I just discovered that payments using saved cards aren’t working on the block-based checkout page, although they work fine on the shortcode-based checkout. It turned out my local setup was showing the shortcode-based checkout even on the block-based page(!), which caused me to miss the issue initially. I spent quite a bit of time debugging, but eventually had to set up a fresh install, which helped uncover the problem. Fortunately, QA hasn’t started yet, so I’ve pushed a fix for this. Could you please take another look when you get a chance?

dkotter (approved):

qasumitbagthariya:
### QA Update  ✅  

-------------

I have verified this PR in the `enhancement/300` branch, which has been fixed and is functioning as intended.

- Simple product
- Subscription product
- Pre-order products
- Block checkout
- Short code checkout

**Pre-order product**

image

**Subscription Product**

image

**Simple Product**

image


**Testing Environment** 

<Details>

- WordPress: 6.8
- Theme: Storefront 4.6.1
- WooCommerce - 9.8.5
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: enhancement/300

</Details>


Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

jeffpaul:
@faisal-alvi I see the following still unchecked in `Next Steps & Considerations` in the linked issue, is that still TBD and thus should hold on this PR or has that been done?

> Ensure compatibility with WooCommerce Subscriptions and other extensions.

faisal-alvi:
@jeffpaul Thanks for pointing that out. I intentionally left that checkbox unchecked for @qasumitbagthariya to cover during his testing. Moreover, this PR does not make extension-specific changes; the update only changes how the payment functions are called, and those changes are applied globally.

@qasumitbagthariya since your testing is complete and includes this aspect, I’ve now checked that point off.

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.8
- Theme: Storefront 4.6.1
- WooCommerce - 9.8.5
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>

Next Step- Ready to Merge 🚀

PR #345, opened 2025-04-25
Fix errors between the Block Checkout and the Shortcode Checkout

mukeshpanchal27:
### All Submissions:

* [x] Does your code follow the [WooCommerce Sniffs](https://github.com/woocommerce/woocommerce-sniffs/) variant of WordPress coding standards?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully run tests with your changes locally?
* [ ] Will this change require new documentation or changes to existing documentation?

---

### Changes proposed in this Pull Request:

Closes #341 .

The issue is related to #240 so i use the base branch `fix/240`. Once it merged it automatically change base branch to trunk.

### Steps to test the changes in this Pull Request:

1. Activate the Square extension.
2. Navigate to WooCommerce > Settings.
3. Go to the Square tab.
4. Access the Advanced Settings section.
5. Set the Debug Mode field to Show on Checkout Page or Both.
6. Now go to shop page and add product in cart.
7. Go to Block Checkout and the Shortcode Checkout.
8. On checkout > Payment options > Select "Credit Card" option
9. Do not fill any details in CC and click on `Place Order` button
10. Check the error in both checkout pages.

#### Shortcode checkout with Detailed error message
Screenshot 2025-04-25 at 9 09 39 AM

#### Block checkout with Detailed error message
Screenshot 2025-04-25 at 9 10 04 AM

#### Shortcode checkout without Detailed error message
Screenshot 2025-04-25 at 9 10 55 AM

#### Block checkout without Detailed error message
Screenshot 2025-04-25 at 9 10 34 AM

### Changelog entry

> Fix - Error message between the Block Checkout and the Shortcode Checkout.

PR #349, opened 2025-05-01
Fix "3D Secure Verification Token is missing" issue

faisal-alvi:
### All Submissions:

---

As reported in the internal Slack thread [here](https://fueled10up.slack.com/archives/C01UCQGPBDJ/p1746077666242469?thread_ts=1745874484.938769&cid=C01UCQGPBDJ), there was a pre-payment validation failure when attempting to use a saved credit card.

After investigation, it was found that the issue began after this commit: https://github.com/woocommerce/woocommerce-square/pull/345/commits/2ba1929198bb6caa9af882f6ae879638954af556. In that update, individual validation checks were moved into a new function, which inadvertently broke the original sequence of checks and led to the issue.

Specifically, the following check must run in between the other validations for the saved card flow to work correctly:
```
if ( Square_Helper::get_post( 'wc-' . $this->get_id_dasherized() . '-payment-token' ) ) {
	return $is_valid;
}
```

Please note that this is targeting the `smoke-testing` branch.

1. Proceed to checkout using a credit card (either saved or a new card).
2. After a successful checkout, add another product to the cart.
3. Try checking out again using the saved card.
4. The payment should be successful without any errors

### Changelog entry

> Fix - 3D Secure Verification Token is missing issue

### DIFF
--- a/includes/Gateway.php
+++ b/includes/Gateway.php
@@ -309,9 +309,16 @@ class Gateway extends Payment_Gateway_Direct {
 		}
 
 		try {
-			$this->get_validation_exception();
+			if ( '' === Square_Helper::get_post( 'wc-' . $this->get_id_dasherized() . '-buyer-verification-token' ) ) {
+				throw new \Exception( '3D Secure Verification Token is missing' );
+			}
+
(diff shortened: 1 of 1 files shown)


faisal-alvi:
Okay, I’ve incorporated the changes into the original PR #345. Closing this one.

PR #382, opened 2025-07-24
Bidirectional Orders Fulfillment Sync between WooCommerce and Square Orders

faisal-alvi:
### All Submissions:

---

Implements bidirectional fulfillment synchronization between WooCommerce and Square orders, enhancing the existing order sync functionality.

#### Key Implementation Details:

- Adds fulfillment details when creating Square orders from WooCommerce orders
- Syncs fulfillment status changes from Square back to WooCommerce orders
- Maintains order status consistency across both platforms
- Leverages Square's OrderFulfillment API for reliable state management

#### Important Notes:

- Only orders processed through Square Payment Gateway are synced to Square
- A polling system is used to fetch Square orders every 15 minutes by default (this interval can be modified using a filter; `wc_square_order_polling_interval_seconds`).
- Webhooks are not used, as they require merchants to create an application in their Square dashboard; a process that may not be straightforward for most users.
- Only the existing Woo orders are eligible for the syncing of fulfillment updates from Square. No new Woo orders are created.
- Only Square order updates that mark an order as either COMPLETED or CANCELLED are synced back to WooCommerce, as those are the only statuses we are targeting for now. When a fulfillment is completed, the corresponding Square order is marked as COMPLETED, which triggers the sync. Similarly, we also track if the order is marked as CANCELLED.
- For refund scenarios, Square creates a new order while WooCommerce deals with the same order, as both platforms have different strategies, refund scenarios are skipped in this initial implementation.
- No new settings added as order sync was partially implemented for the Woo→Square direction already
- No new database tables have been added, and no new button has been introduced to trigger manual sync in this implementation
- This implementation establishes a foundational integration between the platforms, with opportunities for future enhancements

#### Note on Sync Loop Prevention:

The current implementation intentionally avoids syncing WooCommerce order updates back to Square to prevent infinite sync loops. Here’s the potential issue:

1. A WooCommerce order is updated
2. The update syncs to Square
3. The Square polling system detects the change
4. The change syncs back to WooCommerce
5. Steps 1–4 repeat indefinitely

While it’s possible to implement a solution using metadata flags, such approaches may introduce edge cases. It’s important that we carefully evaluate the implications before heading in that direction.

As a result, this implementation includes:
1. Fulfillment data is included when creating a new Square order
2. Fulfillment updates made in Square are synced back to the same WooCommerce order

Closes https://linear.app/a8c/issue/SQUARE-126/implement-two-way-order-sync.
Closes https://linear.app/a8c/issue/SQUARE-98/created-orders-do-not-appear-in-the-merchants-square-dashboard

1. **Test WooCommerce to Square Sync:**
   - Create a new order in WooCommerce using Square Payment Gateway
   - Verify order appears in Square Dashboard with correct fulfillment details
   - Check fulfillment options are visible at the Square order and match with WooCommerce

2. **Test Square to WooCommerce Sync:**
   - Update fulfillment status in Square Dashboard (e.g., mark as delivered/shipped)
   - Verify Square order status is now "completed"
   - Verify WooCommerce order status updates accordingly (this may require waiting for 15 minutes or manually running `wc_square_sync_orders` event; run from the `/wp-admin/admin.php?page=wc-status&tab=action-scheduler&status=pending` page)
   - Check WooCommerce order is set to "completed"

3. **Test with Different Payment Methods:**
   - Check the same process for Credit Card, Cash App, and Digital Wallet Payments.
   - Check order with Gift Card.

4. **Test Other Cases:**
   - Confirm Capture Order functionality remains unaffected and continues to work as expected
   - Confirm proper error handling for failed syncs
   - Check sync behavior with cancelled/refunded orders

### Changelog entry

> Add - Order fulfillment sync between WooCommerce and Square orders.


includes/Handlers/Order_Sync.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Handles syncing orders between WooCommerce and Square.
+ *
+ * @since x.x.x
+ */
+class Order_Sync {
+	public function __construct() {
+		// Initialize hooks.
+		$this->init();
```
Sidsector9: `self::init()`

faisal-alvi: Thanks for the feedback. I've removed the `static` keyword instead.

dkotter (commented):
Overall this looks good to me though I didn't have time to do a super thorough review, pinging in @Sidsector9 to do a deeper review.

Will definitely want to ensure we have tests in place prior to releasing this

includes/Gateway/API.php
```diff
@@ -716,6 +716,78 @@ public function retrieve_gift_card_by_gan( $gan = '' ) {
...
+	 *
+	 * @since x.x.x
+	 * @param array $location_ids Array of location IDs.
+	 * @param string $start_time Start time for the search.
+	 * @param int $limit Number of orders to retrieve.
+	 * @param string $cursor Cursor for pagination.
+	 * @return array Array containing orders and cursor.
+	 * @throws \Exception
+	 */
+	public function search_orders( $location_ids, $start_time, $limit = 100, $cursor = '' ) {
```
Sidsector9:
```suggestion
	public function search_orders( $location_ids = array(), $start_time, $limit = 100, $cursor = '' ) {
```

includes/Gateway/API.php
```diff
@@ -716,6 +716,78 @@ public function retrieve_gift_card_by_gan( $gan = '' ) {
 		return $this->perform_request( $request );
 	}
 
+	/**
+	 * Searches for Square orders.
+	 *
+	 * @since x.x.x
+	 * @param array $location_ids Array of location IDs.
+	 * @param string $start_time Start time for the search.
```
Sidsector9:
```suggestion
	 * @param string $start_time Start time in the RFC 3339 format for the search.
```

includes/Gateway/API.php
```diff
@@ -716,6 +716,78 @@ public function retrieve_gift_card_by_gan( $gan = '' ) {
...
+				foreach ( $errors as $error ) {
+					$error_messages[] = $error->getDetail();
+				}
+				throw new \Exception( 'SearchOrders API failed: ' . implode( ', ', $error_messages ) );
+			}
+		} catch ( \Exception $e ) {
+			wc_square()->log( 'SearchOrders API error: ' . $e->getMessage(), 'error' );
+			throw $e;
+		}
+	}
```
Sidsector9: Let us do this similar to how we implemented other APIs so that the architecture remains consistent across.

This logic can go under `Gateway\API\Requests\Orders.php`.

The response can be set using a response handler, for example:

```php
$this->set_response_handler( \WooCommerce\Square\API\Response::class );
```

And the request can be performed using:

```php
$response = $this->perform_request( $request );
```

The response can then be checked for:

```php
$response->get_data() instanceof \Square\Models\SearchOrdersResponse
```

includes/Gateway/API/Requests/Orders.php
```diff
@@ -169,6 +175,91 @@ function( $line_item ) {
...
+	 * @param \WC_Order            $order        WooCommerce order object.
+	 * @param \Square\Models\Order $order_model Square order object.
+	 * @return \Square\Models\Order
+	 */
+	protected function set_fulfillment_data( \WC_Order $order, \Square\Models\Order $order_model ) {
+		$order_model->setState( 'OPEN' );
+
+		// Create comprehensive fulfillment object.
+		$fulfillment = new \Square\Models\OrderFulfillment( 'PROPOSED' );
+		$fulfillment->setUid( 'woo_order_fulfillment_' . $order->get_id() );
```
Sidsector9: Throughout the plugin, we use `wc_square()->get_idempotency_key()` to set UUID. Any specific reason we are choosing to do it differently here?

includes/Gateway/API/Requests/Orders.php
```diff
@@ -169,6 +175,91 @@ function( $line_item ) {
...
+
+		// Determine fulfillment type based on shipping method.
+		$shipping_methods = $order->get_shipping_methods();
+		$fulfillment_type = empty( $shipping_methods ) ? 'PICKUP' : 'SHIPMENT';
```
Sidsector9: Minor nitpick, allows for easier jump-to navigation.
```php
use Square\Models\FulfillmentType;
$fulfillment_type = empty( $shipping_methods ) ? FulfillmentType::PICKUP : FulfillmentType::SHIPMENT;
```

Sidsector9: Will this set the Fulfillment type to `PICKUP` for Virtual Orders and Downloadable content that don't require shipping?

includes/Gateway/API/Requests/Orders.php
```diff
@@ -169,6 +175,91 @@ function( $line_item ) {
...
+		$fulfillment->setType( $fulfillment_type );
+
+		// Add fulfillment details based on type.
+		if ( 'SHIPMENT' === $fulfillment_type ) {
+			$shipment_details = new \Square\Models\OrderFulfillmentShipmentDetails();
+
+			// Add recipient information.
+			$recipient = new \Square\Models\OrderFulfillmentRecipient();
+			$recipient->setDisplayName( trim( $order->get_shipping_first_name() . ' ' . $order->get_shipping_last_name() ) );
```
Sidsector9:
```suggestion
			$recipient->setDisplayName( trim( $order->get_formatted_shipping_full_name() ) );
```

includes/Handlers/Order_Sync.php
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	/**
+	 * Initialize hooks.
+	 *
+	 * @since x.x.x
+	 */
+	public function init() {
```
Sidsector9, line 25: The `init()` method is redundant if we're going to call it inside `__construct()`. As per best practices, we don't initialize classes and register hooks inside `__construct()`. I'd suggest to call `init()` on the instance of `Order_Sync` instead of calling it inside the constructor.

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+				// Add order note for status change.
+				$wc_order->add_order_note(
+					sprintf(
+						/* translators: %s is the Square state. */
+						__( 'Order status updated from Square. Square state: %s', 'woocommerce-square' ),
+						$square_state
+					)
+				);
+			} else {
+				wc_square()->log(
```
Sidsector9, line 62: Do we have plans in next phases to create a sync status page, like how we have for products? If yes, we should add a `// @todo: move this to a separate order sync status page`

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+			$square_total_amount = $square_total->getAmount() / 100;
+			$current_total       = (float) $wc_order->get_total();
+
+			if ( abs( $current_total - $square_total_amount ) > 0.01 ) { // Allow for rounding differences.
+				// Only update total if order is not paid yet or if it's a refund scenario.
+				if ( ! $wc_order->is_paid() || $square_total_amount < $current_total ) {
+					$wc_order->set_total( $square_total_amount );
+					$updates_made[] = sprintf( 'Total: %s → %s', $current_total, $square_total_amount );
+				}
+			}
```
Sidsector9: Can we use the `Money_Utility` class for this purpose?

```php
$currency             = $square_total->getCurrency();
$square_total_amount = Money_Utility::cents_to_float( $square_total->getAmount(), $currency );
$current_total        = (float) $wc_order->get_total();
$decimals             = Money_Utility::get_currency_decimals( $currency );

$square_total_rounded  = round( $square_total_amount, $decimals );
$current_total_rounded = round( $current_total, $decimals );

if ( $square_total_rounded !== $current_total_rounded ) {
	// Only update total if order is not paid yet or if it's a refund scenario.
	if ( ! $wc_order->is_paid() || $square_total_rounded < $current_total_rounded ) {
		$wc_order->set_total( $square_total_rounded );
		$updates_made[] = sprintf( 'Total: %s → %s', $current_total, $square_total_rounded );
	}
}
```

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+			$wc_order->set_status( 'completed' );
+			$updates_made[] = 'Marked as completed (fulfillment done in Square)';
+
+			$wc_order->add_order_note(
+				__( 'Order marked as completed - fulfillment completed in Square Dashboard/POS.', 'woocommerce-square' )
+			);
+		}
+
+		// 3. Update order total if changed (with safety checks).
+		if ( $square_total ) {
```
Sidsector9, line 82: This will always evaluate to `true` as `getTotalMoney()` always returns an instance of `Square\Models\Money`

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+			} else {
+				wc_square()->log(
+					sprintf(
+						'Status change not allowed: %s → %s for order %d',
+						$wc_status,
+						$new_wc_status,
+						$wc_order->get_id()
+					),
+					'sync'
+				);
```
Sidsector9, line 70: Just to confirm, should we `return` here if status change is not allowed?

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	/**
+	 * Update fulfillment data from Square order.
+	 *
+	 * @since x.x.x
+	 * @param \WC_Order $wc_order WooCommerce order.
+	 * @param \Square\Models\Order $square_order Square order.
+	 */
+	private function update_fulfillment_data( $wc_order, $square_order ) {
```
Sidsector9: This is a minor nitpick based on the observations in the plugin codebase. If we have a function that may or may not do updates, we prefix them with `maybe_`. Since this method can return early without making any updates, we can name it as `maybe_update_fulfillment_data`.

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+				}
+
+				// Add order note if fulfillment completed
+				if ( 'COMPLETED' === $fulfillment_state ) {
+					$note = __( 'Order fulfillment completed in Square.', 'woocommerce-square' );
+					if ( $tracking_number ) {
+						// translators: %s is the tracking number.
+						$note .= sprintf( __( ' Tracking: %s', 'woocommerce-square' ), $tracking_number );
+					}
+					$wc_order->add_order_note( $note );
```
Sidsector9: As this is inside a loop, this will add multiple order notes for a Woo order.
If we really want to add a note for each Square Order Line Item, then should we update the order note to `Order Line Item fulfillment completed in Square.`?

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+				}
+			}
+
+			// Handle pickup details
+			$pickup_details = $fulfillment->getPickupDetails();
+			if ( $pickup_details ) {
```
Sidsector9, line 182: Redundant check as it will always evaluate to `true` as `getPickupDetails()` returns an instance of `Square\Models\FulfillmentPickupDetails`.

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+			$fulfillment_state = $fulfillment->getState();
+			$fulfillment_type  = $fulfillment->getType();
+
+			// Update fulfillment meta
+			$wc_order->update_meta_data( '_square_fulfillment_state', $fulfillment_state );
+			$wc_order->update_meta_data( '_square_fulfillment_type', $fulfillment_type );
+
+			// Handle shipment details
+			$shipment_details = $fulfillment->getShipmentDetails();
+			if ( $shipment_details ) {
```
Sidsector9, line 154: Redundant check as it will always evaluate to `true` as `getShipmentDetails()` returns an instance of `Square\Models\FulfillmentShipmentDetails`.

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+	 * @return \WC_Order|false
+	 */
+	public function find_existing_wc_order_by_square_id( $square_order_id ) {
+		global $wpdb;
+
+		$results = $wpdb->get_results(
+			$wpdb->prepare(
+				"
+				SELECT order_id, meta_key
+				FROM {$wpdb->prefix}wc_orders_meta
```
Sidsector9: This assumes the merchant is already using `HPOS`. Should we consider scenarios where the merchant is still using `wp_post` and `wp_postmeta` tables?

includes/Sync/Order_Importer.php
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	/**
+	 * Find existing WooCommerce order by Square order ID.
+	 *
+	 * @since x.x.x
+	 * @param string $square_order_id Square order ID.
+	 * @return \WC_Order|false
+	 */
+	public function find_existing_wc_order_by_square_id( $square_order_id ) {
```
Sidsector9:
```suggestion
	public function find_existing_wc_order_by_square_order_id( $square_order_id ) {
```

includes/Sync/Order_Polling.php
```diff
@@ -0,0 +1,334 @@
...
+		}
+	}
+
+	/**
+	 * Schedule the polling cron job.
+	 *
+	 * @since x.x.x
+	 */
+	public function schedule_polling() {
+		wp_schedule_event( time(), 'square_order_polling', self::CRON_HOOK );
```
Sidsector9: Any reason why we're not using Action Scheduler the way we're using to sync Products?

includes/Sync/Order_Polling.php
```diff
@@ -0,0 +1,334 @@
...
+			$location_id      = $settings_handler->get_location_id();
+
+			$all_orders  = array();
+			$cursor      = '';
+			$batch_count = 0;
+			$max_batches = 10; // Prevent infinite loops
+
+			do {
+				// Use the API's search_orders method with cursor
+				$response = $api->search_orders( array( $location_id ), $since_time, 100, $cursor );
```
Sidsector9: Do you think we should take the Shifting Window Problem into consideration here?

Since this loop paginates using `since_time` and the cursor, but Square Orders can still be updated during sync, there's a risk that the result set could shift mid-pagination. Specifically, an order that was already processed could get updated from the Square side during the same sync, causing it to move to a later page and be fetched again – leading to duplicate processing.

Should we consider capturing a sync_start_time at the beginning and bounding the query with `updated_at > last_poll_time AND updated_at <= sync_start_time` to avoid both duplication and missing edge cases?

Curious to hear your thoughts!

Sidsector9 (changes requested):

includes/Gateway/API/Requests/Orders.php
```diff
@@ -114,10 +114,16 @@ public function set_update_order_data( \WC_Order $order, \Square\Models\Order $s
 	 *
 	 * @param \WC_Order            $order        WooCommerce order object.
 	 * @param \Square\Models\Order $square_order Square order object.
+	 * @param string $order_request_type The type of order request.
```
iamdharmesh: Minor nitpick:
```suggestion
	 * @param \WC_Order            $order              WooCommerce order object.
	 * @param \Square\Models\Order $square_order       Square order object.
	 * @param string               $order_request_type The type of order request.
```

includes/Gateway/API/Requests/Orders.php
faisal-alvi: Yes, you’re right - assigning `PICKUP` fulfillment to virtual/downloadable products isn’t semantically ideal.

Currently, we’re using `PICKUP` because virtual products don’t require shipping, and this approach ensures such orders are recognized and visible in Square’s system. It’s not a perfect fit, but it maintains functional consistency with our integration requirements.

While skipping fulfillment for virtual products could be more accurate, the current Phase 1 scope requires fulfillment data to be passed for all orders. Additionally, Square mandates fulfillment for orders to be listed in their system (see [Linear#98](https://linear.app/a8c/issue/SQUARE-98/created-orders-do-not-appear-in-the-merchants-square-dashboard)).

So for now, I suggest we retain the current behavior. The use of `PICKUP` for virtual items, though imperfect, ensures compatibility. It might be worth flagging this with the Square team.

I’ll reach out to the Square team to confirm the best approach here and also request support for a more appropriate `DIGITAL` fulfillment type.

includes/Sync/Order_Importer.php
faisal-alvi: Sure, it’s not a required item, but we can definitely consider implementing it in a future phase. For now, adding a todo comment sounds good and won’t hurt, happy to include that.

includes/Sync/Order_Importer.php
faisal-alvi: You're absolutely correct! The condition will always evaluate to `true`, we can check the actual amount value instead.

includes/Sync/Order_Importer.php
faisal-alvi: Good question! No, we should not return early if the status change is not allowed.

The current logic allows for partial updates - even if status change is blocked, other updates might still be valid:
- Fulfillment status updates
- Order total adjustments  
- Timestamp updates
- Fulfillment data sync

For example, a status change might be blocked (e.g., can't go from 'completed' to 'processing'), but fulfillment status updates are still valid and important.

Getting partial updates is better than getting no updates at all, so the current approach of logging the blocked change but continuing with other updates is the right pattern.

includes/Sync/Order_Importer.php
faisal-alvi: You're right that this is inside a loop, but there's an important distinction here:

We're adding one fulfillment per order that targets ALL items in that order. We're not using Square's beta `line_item_application` field to target specific items per fulfillment.

So even if there are multiple fulfillments, each represents a distinct fulfillment action (e.g., different tracking numbers) - not individual line item fulfillments.

The current note "Order fulfillment completed in Square" is accurate because:
- Each fulfillment covers all items in the order
- Each fulfillment has unique details (tracking number, etc.) that should be recorded
- We're not doing line-item-specific fulfillments

If we were using `line_item_application` to target specific items, then "Order Line Item fulfillment completed" would be more appropriate. But since we're targeting all items per fulfillment, the current wording is correct.

We can consider implementing that in future updates, when we begin supporting the sync of orders originally created in Square, as those may include line-item-level fulfillment.

includes/Sync/Order_Polling.php
faisal-alvi: You're right! Using Action Scheduler would be much better than WordPress cron for this use case. Let me do that. Thanks!

includes/Sync/Order_Polling.php
faisal-alvi: Great question about the Shifting Window Problem! You're absolutely right that pagination with time-based queries can cause duplicate processing.

I considered the bounded time window approach but it got complex and heavy in terms of performance, so went with a lightweight solution instead: https://github.com/woocommerce/woocommerce-square/pull/382/commits/0266a469a880d5205bb94fae1d2083a52822bdc1

Reasons:
- Orders are overwritten (not duplicated), so impact is minimal
- 15-minute polling + 100 orders per batch = very rare occurrence
- YAGNI principle - premature optimization

includes/Gateway/API/Requests/Orders.php
Sidsector9: Thanks for the info. I think we should brainstorm and think of the long-term implications of using PICKUP for virtual orders. @iamdharmesh @jeffpaul any thoughts on this?

includes/Sync/Order_Importer.php
faisal-alvi: Actually, `getShipmentDetails()` can return null if the shipment details haven’t been set, so the check is necessary to avoid potential errors.

includes/Sync/Order_Importer.php
faisal-alvi: Actually, `getPickupDetails()` can return null if the shipment details haven’t been set, so the check is necessary to avoid potential errors.

faisal-alvi:
Thanks @Sidsector9 and @iamdharmesh for the helpful suggestions and questions.

- All of them have been addressed! ✅
- Added an E2E test to cover the implementation
- Removed one outdated test related to the now-removed beta Product Editor, per [WooCommerce 10.1 pre-release update](https://developer.woocommerce.com/2025/07/23/10-1-pre-release-updates/)

Requesting a re-review whenever convenient.

includes/Gateway/API/Requests/Orders.php
faisal-alvi: I followed up with the Square team ([on slack](https://fueled.slack.com/archives/C05REAQUK28/p1753794822695979)) and got confirmation from Max@Square:

> Unfortunately, releasing the DIGITAL fulfillment type on the public API is not on this year’s roadmap. I’m working to ensure this request is acknowledged by the product team.
So if using order fulfillments, the PICKUP type will be the best option for now.

So as things stand, continuing with `PICKUP` for virtual/downloadable orders remains the most compatible option for visibility in Square. We can flag this for future revision once Square expands fulfillment support.

includes/Sync/Order_Importer.php
Sidsector9: @faisal-alvi, your previous implementation was correct. I didn't notice it return `?Money` and not `Money`, which means it returns a nullable type.

includes/Sync/Order_Polling.php
Sidsector9: @faisal-alvi this implements dedpulication, but it is still susceptible to the Shifting Window problem. Consider this example:

- You're querying Square with `updated_at > 10:00 AM`
- The sync starts at 10:10 AM (this is your `sync_start_time`).
- Orders are fetched in pages of 3, sorted by `updated_at ASC`.
- Plugin limits itself to `max_batches = 10`

Square orders (sorted by `updated_at`):

| Order ID | updated\_at |
| -------- | ----------- |
| #A       | 10:01 AM    |
| #B       | 10:02 AM    |
| #C       | 10:03 AM    |
| #D       | 10:04 AM    |
| #E       | 10:05 AM    |
| #F       | 10:06 AM    |
| #G       | 10:07 AM    |

#### Sync timeline

| Time         | What Happens                                                                    |
| ------------ | ------------------------------------------------------------------------------- |
| 10:10:01     | Sync starts, captures `sync_start_time = 10:10:01`.                             |
| 10:10:02     | Plugin queries Page 1 with `since=10:00`, gets `[#A, #B, #C]`.                  |
| 10:10:03     | Plugin starts processing `#A`.                                                  |
| 10:10:04     | Plugin processes `#B`.                                                          |
| **10:10:05** | **Square updates `#B`, sets `updated_at = 10:10:08 AM`.**                          |
| 10:10:06     | Plugin processes `#C`.                                                          |
| 10:10:07     | Plugin queries Page 2 using cursor (next page after `#C`). Gets `[#D, #E, #F]`. |
| 10:10:08     | Plugin processes `#D`.                                                          |
| 10:10:09     | Plugin processes `#E`.                                                          |
| 10:10:10     | Plugin processes `#F`.                                                          |
| 10:10:11     | Plugin queries Page 3, gets `[#G]`.                                             |
| 10:10:12     | Plugin processes `#G`.                                                          |
| 10:10:13     | Plugin ends pagination, updates `last_polling_time = 10:10:01`.                 |

Now, when you use `last_polling_time` as the new `sync_start_time` as `10:10:01 AM`, even though `#B` was updated at `10:08 AM`, it never fetches `#B` again because you're fetching all orders after `10:10:01 AM`.

faisal-alvi: Thanks for the explanation @Sidsector9!

I’ve reverted to the lightweight solution and updated the logic to include the end time in the search order query.

You can [click here](https://github.com/woocommerce/woocommerce-square/pull/382/files/36d0ad9a189feb80491a3b21996119bf5816f4a0..e225bc77957c178117e39c2e70ecc34cf260b9a4) to review the latest changes.

faisal-alvi:
@Sidsector9 This is ready for your review following the latest feedback implementation, as noted [here](https://github.com/woocommerce/woocommerce-square/pull/382#discussion_r2260518766).

Sidsector9 (approved):
Thanks for working on those changes @faisal-alvi !

dkotter (approved):

qasumitbagthariya:
### QA Update  ✅ 
-------------

I have verified this PR in the `enhancement/341` branch, which has been fixed and is functioning as intended.

1. Test WooCommerce to Square Sync: ✅

> - Create a new order in WooCommerce using Square Payment Gateway
> - Verify order appears in Square Dashboard with correct fulfillment details
> - Check fulfillment options are visible at the Square order and match with WooCommerce

[Square Dashboard.pdf](https://github.com/user-attachments/files/21833584/Square.Dashboard.pdf)

image



2. Test Square to WooCommerce Sync: ✅

> - Update fulfillment status in Square Dashboard (e.g., mark as delivered/shipped)
> - Verify Square order status is now "completed"
> - Verify WooCommerce order status updates accordingly (this may require waiting for 15 minutes or manually running woocommerce_square_poll_orders event; run from the /wp-admin/admin.php?page=wc-status&tab=action-scheduler&status=pending page)
> - Check WooCommerce order is set to "completed"


https://github.com/user-attachments/assets/96765aed-96cd-4db0-8eb2-734236e4b44b


image


3. Test with Different Payment Methods:

> Check the same process for Credit Card, Cash App, and Digital Wallet Payments.
> Check order with Gift Card.


https://github.com/user-attachments/assets/10ce35c3-861a-49e4-a3b2-10e54d3fd897


https://github.com/user-attachments/assets/cfb6a922-4076-4b6e-b4a7-aae620a781da


4. Test Other Cases:

> Confirm Capture Order functionality remains unaffected and continues to work as expected ✅
> Confirm proper error handling for failed syncs 


**Testing Environment** 

<Details>

- WordPress: 6.8.2
- Theme: Storefront 4.6.1
- Theme: Twenty Twenty-Five 1.2
- WooCommerce - 10.0.4
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: feat/bidirect-order-sync-woo-to-square

</Details>


Testing Document status:

- [x] Yes



@faisal-alvi Other than that I notice below points.

1. I noticed some cancel behavior that looks different. When we cancel the order from the Square dashboard, the WooCommerce status should update to 'Cancelled,' but instead, it appears as 'Completed

https://github.com/user-attachments/assets/a9807661-dd77-4dd8-a4e9-3726b3cc7df0

2. Refund not work from the Square board.

https://github.com/user-attachments/assets/f88e3264-1bc3-4be2-85c3-2e98556416d2

https://github.com/user-attachments/assets/5b03002f-b85f-41fa-afd5-8ad0a1c6f1c4

3. I noticed that when sync fails, the error is logged but not displayed in the edit order section. Just wanted to confirm if this is the expected behavior, or should we also display the error in the edit order section?

image

image

qasumitbagthariya (approved):

faisal-alvi:
@qasumitbagthariya Thanks for the testing!

#### 1. Cancel Behavior

When Square line items are marked as Cancelled in the Square Dashboard:
- The line-item fulfillment status changes to CANCELLED.
- However, the overall Square order status changes to COMPLETED.
- On the next sync cycle, WooCommerce reflects this by marking the entire order as Completed, since WooCommerce doesn’t support line-item–level fulfillment statuses, only an overall order status. cc @dkotter

#### 2. Refunds from Square Dashboard

The refund behavior in the Square Dashboard is not in our control and may also be related to using a test card in the sandbox account. Phase 1 is focused only on fulfillment status sync across both platforms, other things, including refund handling, can be considered as part of future enhancements if needed.

#### 3. Error Logging vs Display in Edit Order

During our on-call testing, we added code to intentionally trigger an error:

```
// force error to trigger the log
throw new \Exception( 'Test error' );
```

This was mistakenly placed above:

```
$existing_order = $importer->find_existing_wc_order_by_square_order_id( $order_id );
```

Because of this, `$existing_order` was `null`, which prevented the error note from being added to the order. If placed correctly, the error would have been logged and shown in the Edit Order section as intended.

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.8.2
- Theme: Twenty Twenty-Five 1.3
- WooCommerce - 10.1.0
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>

Next Step- Ready to Merge 🚀

PR #392, opened 2025-08-11
Accurately count attribute values – 250 values, not chars

faisal-alvi:
### All Submissions:

---

**Fix attribute validation to count values instead of string length**

The WooCommerce Square plugin was incorrectly validating attribute values by checking the character length of pipe-separated strings instead of counting the actual number of attribute values. This caused validation failures for products with long attribute value names, even when they had fewer than 250 values, while allowing products with many short attribute values to pass validation even when exceeding the 250 limit.

This PR modifies `validateMaxAttributeValues()` function to count array values directly instead of converting to a string first

Closes https://linear.app/a8c/issue/SQUARE-147/attribute-limitation-is-250-values-but-plugin-js-checks-string-not.

1. Create a product with one attribute containing 6 values, each with 50+ characters (e.g., "Very long attribute value name that exceeds typical length for testing purposes")
2. Press tab while editing; this will trigger a values check, confirm no errors are displayed. Previously, this would fail with "Too many attribute values" error.
3. Verify the product now syncs successfully since it has only 6 values (under the 250 limit)
4. Test with a product that has 251+ attribute values to confirm validation still works correctly and error is displayed (you can do this just by copy-pasting any string (for example "Red | ")
5. Verify error message

### Changelog entry

> Fix - Accurately count attribute values – 250 values, not characters.

### DIFF
--- a/src/js/admin/wc-square-admin-products.js
+++ b/src/js/admin/wc-square-admin-products.js
@@ -817,7 +817,7 @@ jQuery( document ).ready( ( $ ) => {
 			);
 		};
 
-	const validateMaxAttributeValues = ($field) => {
+		const validateMaxAttributeValues = ($field) => {
 			let attrValues = $field.val();
 
 			// Count the actual number of values.
@@ -833,7 +833,7 @@ jQuery( document ).ready( ( $ ) => {
 				valueCount,
 				250,
 				$field,
-				wc_square_admin_products.i18n.too_many_attribute_values + ': ' + valueCount + ' values (max 250)'
+				wc_square_admin_products.i18n.too_many_attribute_values.replace( '%d', valueCount )
 			);
 		};
 
--- a/src/js/admin/wc-square-admin-products.js
+++ b/src/js/admin/wc-square-admin-products.js
@@ -817,19 +817,23 @@ jQuery( document ).ready( ( $ ) => {
 			);
 		};
 
-		const validateMaxAttributeValues = ($field) => {
+	const validateMaxAttributeValues = ($field) => {
 			let attrValues = $field.val();
 
-			// if attrValues is array (when taxonomy attribute), convert to string and combine values by '|'.
+			// Count the actual number of values.
+			let valueCount;
 			if (Array.isArray(attrValues)) {
-				attrValues = attrValues.join('|');
+				valueCount = attrValues.length;
+			} else {
+				// If it's already a string (pipe-separated), split and count.
+				valueCount = attrValues.split('|').filter(value => value.trim() !== '').length;
 			}
 
 			return validateLimit(
-				attrValues.length,
+				valueCount,
 				250,
 				$field,
-				wc_square_admin_products.i18n.too_many_attribute_values + ': ' + attrValues.substring(0, 50) + '...'
+				wc_square_admin_products.i18n.too_many_attribute_values + ': ' + valueCount + ' values (max 250)'
 			);
 		};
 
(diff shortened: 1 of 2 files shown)


src/js/admin/wc-square-admin-products.js
```diff
@@ -817,19 +817,23 @@ jQuery( document ).ready( ( $ ) => {
...
+				valueCount = attrValues.split('|').filter(value => value.trim() !== '').length;
 			}
 
 			return validateLimit(
-				attrValues.length,
+				valueCount,
 				250,
 				$field,
-				wc_square_admin_products.i18n.too_many_attribute_values + ': ' + attrValues.substring(0, 50) + '...'
+				wc_square_admin_products.i18n.too_many_attribute_values + ': ' + valueCount + ' values (max 250)'
```
dkotter: If we hardcode ` values (max 250)` here, this won't be translateable

src/js/admin/wc-square-admin-products.js
```diff
@@ -817,19 +817,23 @@ jQuery( document ).ready( ( $ ) => {
 			);
 		};
 
-		const validateMaxAttributeValues = ($field) => {
+	const validateMaxAttributeValues = ($field) => {
```
dkotter: Do we have similar validation on the server-side? Or do we only validate this limit on the client-side? Just wondering if 1 - we need to apply this same fix to the server-side or 2 - add server-side validation if it doesn't exist

dkotter (changes requested):

src/js/admin/wc-square-admin-products.js
faisal-alvi: We don't have server-side validation because Square's API handles the real enforcement. The client-side validation just provides immediate feedback to merchants on the product edit page.  We can keep the current approach (client-side validation only) unless there's a specific business need for server-side validation. What's your suggestion; should we add server-side validation when products are saved/updated, or is the current client-side + API failure approach sufficient?

dkotter (approved):

qasumitbagthariya:
### QA Update  ✅ 
-------------

I have verified this PR in the `fix/147-accurate-attribute-count-250-values` branch, which has been fixed and is functioning as intended.

- Test with a product that has 251+ attribute values to confirm validation still works correctly and error is displayed  ✅ 

**Before**

image

**After**
image


**Testing Environment** 

<Details>

- WordPress: 6.8.2
- Theme: Twenty Twenty-Five 1.2
- WooCommerce - 10.1.0
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>


Testing Document status:

- [x] Yes





@faisal-alvi @dkotter 

**Suggestion:**
It seems like most of the limitations are from the Square side, which we can handle in a similar way. For example:

1. If an attribute exceeds 65 characters, we should show an error at the time of entry itself, so that it doesn’t cause issues during sync.

qasumitbagthariya (approved):

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.8.2
- Theme: Twenty Twenty-Five 1.3
- WooCommerce - 10.1.0
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>


Next Step- Ready to Merge 🚀

PR #440, opened 2025-12-17
[Phase 1] Add Square Discount Codes (Coupons) Support

peterwilsoncc:
### All Submissions:

---

Closes https://linear.app/a8c/issue/SQUARE-219/implement-discount-codes

This PR adds support for Square discount codes (coupons) in WooCommerce Square: customers can apply Square-created discount codes on the cart and at checkout. Discounts are calculated by Square via the `CalculateOrder` API, stored in session, and applied through WooCommerce’s coupon flow so cart/order totals and display stay correct. Square discount codes and WooCommerce coupons are mutually exclusive (only one type can be applied at a time).

---

**Woo Order Example Screenshot:**

image

---

**Square Order Example Screenshot:**

image

---

- **Square discount codes (coupons):** Customers can apply Square discount codes at cart and checkout. Discounts are calculated via Square’s `CalculateOrder` API with `proposed_discount_codes`; per-line-item and total discount are stored in session and applied through WooCommerce’s coupon flow. Square and WooCommerce coupons are mutually exclusive (one type at a time).
- **CalculateOrder:** Handled as a special case in `do_square_request()` using direct HTTP so we can send `proposed_discount_codes`; the Square PHP SDK’s CalculateOrderRequest only supports `order` and `proposed_rewards` (loyalty).
- **Gateway:** After creating the Square order, the plugin creates a Redemption to link the Square discount code to the Square order. WC discount line items are omitted when a Square discount code is present (discount applied via Redemption). Order meta stores `_square_discount_code_id`, `_square_discount_amount`, `_square_discount_per_item`, and `_square_redemption_id`.
- **Zero-total:** If a Square discount would make the order total zero or less, the coupon is rejected, removed from the cart, Square session data is cleared, and an error notice is shown (Square cannot process zero-amount transactions).
- **Connection:** `DISCOUNT_CODES_READ` and `DISCOUNT_CODES_WRITE` scopes added for discount code search and redemption.

Added a setting field to control the feature:

image

Notes:
- The Square PHP SDK’s `CalculateOrderRequest` does not support `proposed_discount_codes` parameter. The plugin uses direct HTTP for the CalculateOrder call so discount code IDs can be sent while keeping the existing request/response pattern.
- Out of scope for this phase: We confirmed with Square that “Free Shipping” / “Free Delivery” in Square Online are handled client-side on service charges and are not part of the discount/line-item APIs, so we are treating them as out of scope for this phase and focusing on product/line-item discount codes only.
- The Coupon API is not available in the Square sandbox (we have to use a production account), so E2E tests will be converted into a separate PR once the Coupon API is available in the sandbox. A linear ticket created: https://linear.app/a8c/issue/WOO10UP-43/e2e-square-discount-codes-coupons-add-tests-when-coupon-api-available

---

**Note:**
This is a significant enhancement. Please ensure that various checkout flows are tested thoroughly, including (but not limited to):

- Prices exclusive of tax with taxable shipping charges
- Non-taxable shipping charges
- No shipping charges
- Multiple coupons applied
- Multiple tax rates (percentage-based)
- Combination of multiple percentage-based coupons and multiple fixed-amount coupons
- No coupons applied
- WooCommerce core coupons applied
- Verification that tax amounts, discounts, and order totals match correctly on both platforms

1. (in trunk) Connect your site with the Square if not already.
2. Move to a fix branch.
3. Check the new setting field to enable the discount code feature, which should be enabled by default.
4. See the admin notice that you need to reconnect your Square account because of a lack of required scopes.
5. Reconnect WooCommerce Square with a Square account that has DISCOUNT_CODES_READ and DISCOUNT_CODES_WRITE (reconnect if needed to grant new scopes).
6. In Square Dashboard, create a discount code (e.g. 10% off or fixed amount) and note the code.
7. On the storefront, add a product, enter the Square discount code, and apply; confirm the discount appears and totals update.
8. Update qty if required; confirm discount recalculates and persists.
9. Complete payment with Square; confirm the order is created and, in Square, the order shows the discount/redemption.
10. Apply a Square code that would make the order total zero (e.g. 100% off); confirm the coupon does not apply, an error notice is shown, and the cart total is unchanged.
11. With a Square code applied, try to apply a WooCommerce coupon; confirm it is rejected with a notice. Remove the Square code and apply a WC coupon; then try to apply a Square code and confirm it is rejected.
12. Create an order without any discount code; confirm checkout and payment still work and no redemption is created.
13. Most importantly, try as many different coupons and various cases and confirm the checkout flow works with correct numbers.
14. Use the new filter `woocommerce_square_enable_discount_codes` to opt-out from Square coupons from the site. Add this code to your `functions.php`
```
// In theme's functions.php or a custom plugin
add_filter( 'woocommerce_square_enable_discount_codes', '__return_false' );
```
---

### Changelog entry

> Add - [Phase 1] Add Square Discount Codes (Coupons) Support


includes/API.php
```diff
@@ -1076,7 +1084,163 @@ protected function do_post_parse_response_validation() {
...
+			if ( isset( $calculated_order_data['net_amounts'] ) && isset( $calculated_order_data['net_amounts']['total_money'] ) ) {
+				$net_amounts = new \Square\Models\OrderMoneyAmounts();
+				$net_total   = new \Square\Models\Money();
+				if ( isset( $calculated_order_data['net_amounts']['total_money']['amount'] ) ) {
+					$net_total->setAmount( $calculated_order_data['net_amounts']['total_money']['amount'] );
+				}
+				if ( isset( $calculated_order_data['net_amounts']['total_money']['currency'] ) ) {
+					$net_total->setCurrency( $calculated_order_data['net_amounts']['total_money']['currency'] );
+				} else {
+					$net_total->setCurrency( $square_order->getTotalMoney() ? $square_order->getTotalMoney()->getCurrency() : 'USD' );
```
iamdharmesh: Nit:
Same as earlier could we use currency from Woo order or at least store's currency as a fallback?

includes/Gateway/API/Requests/Orders.php
```diff
@@ -367,6 +440,24 @@ protected function get_api_line_items( \WC_Order $order, $line_items, $taxes ) {
...
+				$coupon_code             = $applied_coupons[0];
+				$square_discount_code_id = WC()->session->get( '_square_discount_code_id_' . $coupon_code );
+
+				// If found in session, store it in order meta for future use.
+				if ( ! empty( $square_discount_code_id ) ) {
+					$order->update_meta_data( '_square_discount_code_id', $square_discount_code_id );
+					$order->save_meta_data();
+				}
+			}
+		}
```
iamdharmesh: Noticed similar fallback code in the `includes/Gateway.php` as well maybe better to move this to the utility function or maybe add this to `woocommerce_checkout_update_order_meta` hook to make sure that metadata always there?

includes/Gateway/API.php
```diff
@@ -463,6 +495,97 @@ public function update_order( \WC_Order $order, \Square\Models\Order $square_ord
...
+			array(
+				'headers' => array(
+					'Authorization'  => 'Bearer ' . $bearer_token,
+					'Content-Type'   => 'application/json',
+					'Square-Version' => '2025-01-23',
+				),
+				'body'    => wp_json_encode( $request_body ),
+				'timeout' => 30,
+			)
+		);
```
iamdharmesh: Nit:
It seems we are getting the token and calling API via wp_remote_post in the other places as well. How about create a wrapper function which accept API endpoint/path and body to call Square API.

includes/API.php
```diff
@@ -1076,7 +1084,163 @@ protected function do_post_parse_response_validation() {
...
+				array(
+					'headers' => array(
+						'Authorization'  => 'Bearer ' . $bearer_token,
+						'Content-Type'   => 'application/json',
+						'Square-Version' => '2025-01-23',
+					),
+					'body'    => wp_json_encode( $request_body ),
+					'timeout' => 30,
+				)
+			);
```
iamdharmesh: Nit: Same as earlier, wrapper function could be used here.

includes/API.php
```diff
@@ -1076,7 +1084,163 @@ protected function do_post_parse_response_validation() {
...
+			if ( isset( $calculated_order_data['total_money'] ) ) {
+				$total_money = new \Square\Models\Money();
+				if ( isset( $calculated_order_data['total_money']['amount'] ) ) {
+					$total_money->setAmount( $calculated_order_data['total_money']['amount'] );
+				}
+				if ( isset( $calculated_order_data['total_money']['currency'] ) ) {
+					$total_money->setCurrency( $calculated_order_data['total_money']['currency'] );
+				} else {
+					// Fallback to original order currency.
+					$total_money->setCurrency( $square_order->getTotalMoney() ? $square_order->getTotalMoney()->getCurrency() : 'USD' );
```
iamdharmesh: Nit:
Instead of using hardcode `USD` as a fallback, could we use currency from Woo order or at least store's currency?

includes/Coupons.php
```diff
@@ -0,0 +1,1009 @@
...
+	}
+
+	/**
+	 * Prevent non-Square coupons from being used with Square coupons.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param bool      $is_valid Whether the coupon is valid.
+	 * @param \WC_Coupon $coupon  Coupon object.
+	 * @param \WC_Discounts $discounts Discounts object.
```
iamdharmesh: Let's remove this as function only accept 2 params.
```suggestion
```

iamdharmesh (commented):
Thanks for the PR, @faisal-alvi great work here.

I haven’t completed the full review yet, but I’ve noticed a few things I’d like to list here so you can take a look while I finish the review.

1. After applying a Square coupon/discount code, I’m still able to place the order using another payment gateway, which shouldn’t happen. Could you please check this?

2. The discount does not update correctly after adding a new product to the cart or updating the product quantity. Sometimes it works, sometimes it doesn’t. Please refer to the screenshot below. the 10% discount is applied, but the discount amount is not updated correctly after reducing the quantity.
image

3. When a Square coupon code is applied and we then try to apply a WooCommerce coupon code, it shows “Coupon is not valid.” It would be better to provide a clearer message indicating that both codes cannot be used together, if possible.
image


**Questions:**
* Do we have any option on the WooCommerce side to enable or disable Square discount codes on the site?
* With the new `DISCOUNT_CODES_READ` and `DISCOUNT_CODES_WRITE` scopes, users will need to reconnect their Square account to benefit from this functionality. If a user does not grant these new scope permissions, will anything break in the WooCommerce coupon application flow? It’s fine if they’re unable to use this new feature, but the WooCommerce core coupon flow should continue to work without issues.

Thank you.

includes/Utilities/Coupon_Utility.php
```diff
@@ -0,0 +1,393 @@
...
+		// Create cache key based on pricing rule ID and version.
+		$pricing_rule_id      = $square_discount_code['pricing_rule_id'];
+		$pricing_rule_version = $square_discount_code['pricing_rule_version'];
+		$cache_key            = 'square_coupon_mapping_' . md5( $pricing_rule_id . '_' . $pricing_rule_version );
+
+		// Check cache first.
+		$cached = get_transient( $cache_key );
+		if ( false !== $cached && is_array( $cached ) ) {
+			// Update code in case it's different (though it should be the same for same pricing rule).
+			$cached['code'] = $square_discount_code['code'];
```
iamdharmesh, line 515: Not very much sure on the how discount code is connected with pricing rule is there 1:1 relation between them? Could you please provide bit context on it?

includes/Gateway.php
```diff
@@ -486,6 +486,38 @@ protected function do_transaction( $order ) {
...
+							$error_message = is_wp_error( $redemption_result ) ? $redemption_result->get_error_message() : __( 'Unknown error', 'woocommerce-square' );
+							// translators: %1$s: discount code ID, %2$s: order ID, %3$s: error message.
+							$this->get_plugin()->log( sprintf( 'Square: Failed to create redemption for discount code %1$s on order #%2$s: %3$s', $square_discount_code_id, $order->get_id(), $error_message ), $this->get_id() );
+						}
+					} catch ( \Exception $redemption_exception ) {
+						// Log redemption errors but don't fail the payment.
+						if ( $this->debug_log() ) {
+							// translators: %1$s: discount code ID, %2$s: order ID, %3$s: error message.
+							$this->get_plugin()->log( sprintf( 'Square: Error creating redemption for discount code %1$s on order #%2$s: %3$s', $square_discount_code_id, $order->get_id(), $redemption_exception->getMessage() ), $this->get_id() );
+						}
```
iamdharmesh: Just curious if we should fail the order here. What are the reasons for the create redemption failure? if coupon is already reached to redemption limit then will this fail? Could you please help on this?

Consider example we have 1 coupon which has limit on how many times it can be redeemed (`Limit number of times voucher can be redeemed` option). Now, I suspect if redemption limit reached it will fail the create redemption here. As we are ignoring error here, will discount still apply to the order payment?

includes/Coupons.php
```diff
@@ -0,0 +1,1203 @@
...
+		// Check if there are any applied coupons.
+		if ( ! empty( $applied_coupons ) ) {
+			$has_square_coupon = ! empty( self::get_applied_square_coupon_codes() );
+
+			// If trying to apply Square coupon but WooCommerce coupon exists.
+			if ( $is_square_coupon && ! $has_square_coupon ) {
+				$coupon->set_error_message(
+					sprintf(
+						/* translators: %s: coupon code */
+						__( 'Sorry, the Square discount code "%s" cannot be used in combination with WooCommerce coupons. Please remove the WooCommerce coupon and try again.', 'woocommerce-square' ),
```
iamdharmesh: Nit: End user (customer) might not be aware of which platform is used to create this store. So, this error message might confuse them or will not help them much. We need to update this to something more generic. Maybe this could be raised in the internal UAT and get this updated as per conclusion there.

includes/Coupons.php
```diff
@@ -0,0 +1,1203 @@
...
+	 *
+	 * @param string     $discount_code The discount code.
+	 * @param array|null $code_details  The discount code details to cache. Null if not found.
+	 */
+	public static function set_cache_discount_code( $discount_code, $code_details ) {
+		$transient_key = self::get_discount_code_transient_key( $discount_code );
+
+		// Cache for 1 hour. Store false if not found to distinguish from "not cached yet".
+		$cache_value = null === $code_details ? false : $code_details;
+		set_transient( $transient_key, $cache_value, HOUR_IN_SECONDS * 1 );
```
iamdharmesh: It seems on project description it is mentioned that coupon code should be cached for 15 mins, while it's 1 hour here. So, just wanted to check if this is decided to update it to 1 hour.

> Square coupon look-ups (known and unknown) will be cached for 15 minutes. Revalidation of coupons in the cache for greater than three minutes will take place during the check out process.

iamdharmesh (commented):
Thanks for the changes @faisal-alvi. Noticed few more things.

1. It seems discount evenly distributed amongst coupon code instead of actual discount amount. Could you please help to check once. In this video I have applied 2 copouns 1. 5 FIXED = $5 fixed, ANNUALSALE1 = 10% off.

https://github.com/user-attachments/assets/b5651e0b-05e2-49e9-9a2a-7062885e7098

2. For coupon limited to specific product gives proper error when no coupon code applied but it don't throw any error when there is already coupon code applied.

https://github.com/user-attachments/assets/6ecdf4a9-0d60-4af1-9017-1074914a65e7

Thanks.

includes/Utilities/Coupon_Utility.php
faisal-alvi: It's not a 1:1 relation but a many:1. In Square, several discount codes (e.g. SAVE20, SAVE20BACKUP) can point to the same pricing rule. The rule (and its related discount) defines the behavior (type, amount, product set); the code is just the string the customer enters.

We cache a pricing rule (discount_type, amount, product_ids). So the cached value is per pricing rule, but the code string is per discount code. We overwrite code with the current discount code so the returned WC coupon has the correct code (e.g. the one the customer entered), even when the mapping came from cache.

includes/Gateway.php
faisal-alvi: Great catch! You're right to be concerned. If a coupon has reached its redemption limit, it won't be applied in the first place when trying to add it on the cart or checkout page. However, if somehow due to specific coupon limitations it doesn't get recognized at that point, we should fail the order on redemption failure. I've updated the codebase to throw an error and prevent the payment from proceeding when redemption fails.

includes/Coupons.php
faisal-alvi: That wasn't an intentional change. The project description is correct: coupon lookups should be cached for 15 minutes. The 1-hour value was added for better performance. The 15-minute option works either way. I've updated the implementation so the discount code cache TTL is now 15 minutes to match the spec. Thanks for flagging this!

faisal-alvi:
@iamdharmesh Thanks for the feedback. I've addressed your points- could you please give another look and let me know if anything needs adjustment?

> Do we have any option on the WooCommerce side to enable or disable Square discount codes on the site?

We don't currently have a UI option for this, as it wasn't part of the original requirements. I've added a filter `woocommerce_square_enable_discount_codes` that allows users to opt out programmatically, but it's worth discussing with the Square team whether we should add a dedicated settings field in the admin area for this.

> With the new DISCOUNT_CODES_READ and DISCOUNT_CODES_WRITE scopes, users will need to reconnect their Square account to benefit from this functionality. If a user does not grant these new scope permissions, will anything break in the WooCommerce coupon application flow? It’s fine if they’re unable to use this new feature, but the WooCommerce core coupon flow should continue to work without issues.

Yes, the WooCommerce core coupon flow will continue to work without issues even if users don't grant the new DISCOUNT_CODES_READ and DISCOUNT_CODES_WRITE scope permissions. If the API call fails (e.g. 403 from missing scopes), the code returns `null`. The code is then treated as not a Square discount code and is handled as a normal WooCommerce coupon.

includes/Gateway.php
```diff
@@ -479,40 +479,99 @@ protected function do_transaction( $order ) {
...
+									/* translators: %s: exception message */
+									__( 'The coupon could not be applied. Please remove it and try again, or use a different payment method. Error: %s', 'woocommerce-square' ),
+									$redemption_exception->getMessage()
+								)
+							);
+						}
+					}
+					if ( ! empty( $redemption_ids ) ) {
+						$order->update_meta_data( '_square_redemption_ids', $redemption_ids );
+						$order->update_meta_data( '_square_redemption_id', $redemption_ids[0] );
```
iamdharmesh: Just curious, any specific reason to store the first redemption_id separately?

includes/Coupons.php
```diff
@@ -0,0 +1,1397 @@
...
+			// Note: We do NOT add WooCommerce discounts here - Square will calculate via discount codes.
+			$line_items[] = $line_item;
+		}
+
+		// Intentionally do NOT add shipping to the order sent to CalculateOrder.
+		// When shipping is included, Square prorates/allocates discounts across product + shipping,
+		// so fixed/capped coupons show less on product (e.g. $5 fixed becomes $4.82). Excluding
+		// shipping keeps discount amounts consistent regardless of shipping method. The cart/checkout
+		// total still adds shipping via WooCommerce; at payment we send the full order (with shipping)
+		// and use adjust_order so the charge matches the displayed total.
```
iamdharmesh: Adjust order may used in edge/rare case and we may don't want to use it with every order, do we have any way in which we don't have to use the adjust order?

includes/Coupons.php
```diff
@@ -0,0 +1,1397 @@
...
+				}
+			}
+		}
+
+		if ( empty( $square_discount_code_ids ) ) {
+			return false;
+		}
+
+		$order->update_meta_data( '_square_discount_code_ids', $square_discount_code_ids );
+		$order->update_meta_data( '_square_discount_code_id', $first_id );
```
iamdharmesh: Just curious, any reason for saving this first ID separately?

iamdharmesh (approved):
Thanks for the changes @faisal-alvi. Added few notes to check but pre-approving PR to unblock QA on this. 

I didn't yet tested and compared what we see on WooCommerce side and what get saved on Square side for each case but could you please check on that adjust order thing we discussed yesterday.

> Yes, the WooCommerce core coupon flow will continue to work without issues even if users don't grant the new DISCOUNT_CODES_READ and DISCOUNT_CODES_WRITE scope permissions. If the API call fails (e.g. 403 from missing scopes), the code returns null. The code is then treated as not a Square discount code and is handled as a normal WooCommerce coupon.

Thanks for information. In this case we may don't want to call API if user don't have DISCOUNT_CODES scope. We can check it using [retrieve token](https://developer.squareup.com/reference/square/o-auth-api/retrieve-token-status) API. This is connected with how user onboard to the feature so, better to get this discussed in UAT call.

includes/Coupons.php
faisal-alvi: Correct, this has been improved in the latest changes. We now only adjust the order totals in cents, which is expected because Square and WooCommerce use different rounding methods. We have been doing this for a long time; this is not new in this PR.

includes/Gateway.php
faisal-alvi: Nice catch! It was there initially, but I removed it after improving the functionality to support multiple redemptions.

includes/Gateway/API/Requests/Orders.php
```diff
@@ -365,7 +514,12 @@ protected function get_shipping_line_items( \WC_Order $order ) {
 	 */
 	protected function get_api_line_items( \WC_Order $order, $line_items, $taxes ) {
 		$api_line_items = array();
-		$tax_type       = wc_prices_include_tax() ? API::TAX_TYPE_INCLUSIVE : API::TAX_TYPE_ADDITIVE;
+		// Always use ADDITIVE with ex-tax base so Square applies tax once and totals match WooCommerce.
+		$tax_type = API::TAX_TYPE_ADDITIVE;
+
+		// Square discount code IDs are set via woocommerce_checkout_create_order and woocommerce_checkout_update_order_meta.
+		$square_discount_code_ids = \WooCommerce\Square\Coupons::get_order_square_discount_code_ids( $order );
+		$has_square_discount      = ! empty( $square_discount_code_ids );
```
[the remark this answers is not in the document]
iamdharmesh: @faisal-alvi This seems bigger change I think it should respect the `wc_prices_include_tax()`.

includes/Gateway/API/Requests/Orders.php
```diff
@@ -485,8 +643,10 @@ protected function get_api_line_items( \WC_Order $order, $line_items, $taxes ) {
 	 * @return \Square\Models\OrderLineItemTax[]
 	 */
 	protected function get_order_taxes( \WC_Order $order ) {
-		$taxes    = array();
-		$tax_type = wc_prices_include_tax() ? API::TAX_TYPE_INCLUSIVE : API::TAX_TYPE_ADDITIVE;
+		$taxes = array();
+
+		// Align with get_api_line_items: always ADDITIVE so order-level tax and line item tax match.
+		$tax_type = API::TAX_TYPE_ADDITIVE;
```
iamdharmesh: Same applies here,

includes/Gateway/Cash_App_Pay_Gateway.php
```diff
@@ -302,6 +302,10 @@ public function init_form_fields() {
 	 * @return bool
 	 */
 	public function is_available() {
+		// Hide Cash App when a Square coupon is applied. This can be enabled with future enhancements.
+		if ( \WooCommerce\Square\Coupons::cart_has_square_coupon() ) {
+			return false;
+		}
```
iamdharmesh, line 308: Maybe something we want to include in the public facing docs.

includes/Utilities/Token_Scope_Utility.php
```diff
@@ -0,0 +1,155 @@
...
+			$url,
+			array(
+				'headers' => array(
+					'Authorization'  => 'Bearer ' . $access_token,
+					'Content-Type'   => 'application/json',
+					'Square-Version' => '2026-01-22',
+				),
+				'timeout' => 15,
+			)
+		);
```
iamdharmesh, line 93: Just curious if we have any method at SDK to get the scopes and we can directly use it.

includes/Coupons.php
```diff
@@ -0,0 +1,1199 @@
...
+			}
+		}
+
+		foreach ( $coupon_codes_in_calc as $code ) {
+			WC()->session->set( '_square_discount_code_id_' . $code, Coupon_Utility::get_square_discount_code_id_by_code( $code ) );
+
+			$amount   = $per_coupon_amount[ $code ] ?? 0;
+			$per_item = $per_coupon_per_item[ $code ] ?? array();
+			$is_fixed = isset( $code_to_discount_type[ $code ] ) && in_array( $code_to_discount_type[ $code ], array( 'fixed_cart', 'fixed_product' ), true );
+			if ( $is_fixed && 1.0 !== $inclusive_multiplier ) {
```
iamdharmesh: Maybe better to add some code comments here as well for why we are adding tax on the coupons.

includes/Gateway/Cash_App_Pay_Gateway.php
```diff
@@ -1093,20 +1097,23 @@ protected function do_transaction( $order ) {
...
-				$wc_total     = Money_Utility::amount_to_cents( $order->get_total() );
-				$square_total = $response->getTotalMoney()->getAmount();
-				$delta_total  = $wc_total - $square_total;
+				// Adjust order by delta so Square total matches WooCommerce displayed total (charge what the customer saw).
+				// When Square redemption is used, pass current order so adjustment is a service charge (not discounted again); otherwise line item.
+				$square_discount_code_ids = \WooCommerce\Square\Coupons::get_order_square_discount_code_ids( $order );
+				$square_coupon_in_use     = ! empty( $square_discount_code_ids ) ? $response : null;
+				$wc_total                 = Money_Utility::amount_to_cents( $order->get_total() );
+				$square_total             = $response->getTotalMoney()->getAmount();
+				$delta_total              = $wc_total - $square_total;
```
iamdharmesh: Is seems we are removing the cash app pay from the list and here we are handling the discount codes, maybe this is accidently pushed?

includes/Gateway.php
```diff
@@ -479,40 +480,98 @@ protected function do_transaction( $order ) {
...
+										__( 'The coupon could not be applied (e.g. redemption limit reached or code expired). Please remove it and try again, or use a different payment method. Error: %s', 'woocommerce-square' ),
+										$error_message
+									)
+								);
+							}
+						} catch ( \Exception $redemption_exception ) {
+							if ( $this->debug_log() ) {
+								$this->get_plugin()->log( sprintf( 'Square: Error creating redemption for discount code %1$s on order #%2$s: %3$s', $square_discount_code_id, $order->get_id(), $redemption_exception->getMessage() ), $this->get_id() );
+							}
+							throw new \Exception(
```
iamdharmesh: It seems 2 `throw new \Exception(` here, do we needed both?

iamdharmesh (changes requested):
Hi @faisal-alvi, Thanks for the changes. I have added some feedback and some feedback are added by Copilot as well and worth it to go through it. Thanks.

includes/Utilities/Token_Scope_Utility.php
faisal-alvi: Yes, the Square PHP SDK does not expose the OAuth2 token status endpoint (RetrieveTokenStatus).

includes/Gateway/Cash_App_Pay_Gateway.php
faisal-alvi: Thanks. This has been added in [the doc](https://docs.google.com/document/d/15B0Ma1R7rpqTGh-QNNuPBvLYwUH456UxVNws3qHvOyw/edit?tab=t.0) under "Not included in the first phase" title.

includes/Gateway/API/Requests/Orders.php
faisal-alvi: Thanks for flagging this. The ADDITIVE setting with an ex-tax base was intentional — I've tested it across multiple scenarios (with/without Square discount codes, inclusive and exclusive tax) and it holds up well. A thorough review by @qasumitbagthariya, as outlined in the PR description, should be sufficient to keep it as-is.

includes/Gateway/API/Requests/Orders.php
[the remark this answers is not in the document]
faisal-alvi: 👍🏻, add a reply at https://github.com/woocommerce/woocommerce-square/pull/440#discussion_r2826714917

includes/Gateway/API/Requests/Orders.php
iamdharmesh: @faisal-alvi as per our internal discussion you are doing for prevent unnecessary adjustment due to rounding tax. But I am bit concerned for the any regression this can cause. Could you please apply this change conditionally for the new coupons calculations only and not for the default setup? By that way we can make sure that we are not introducing any regression. Thanks.

includes/Utilities/Coupon_Utility.php
```diff
@@ -0,0 +1,687 @@
...
+	 * @param array  $discount_codes Array of discount codes from API response.
+	 * @param string $coupon_code    The coupon code to find and validate.
+	 * @return array|null The matching and valid discount code, or null if not found or invalid.
+	 */
+	private static function find_valid_discount_code( $discount_codes, $coupon_code ) {
+		if ( empty( $discount_codes ) || ! is_array( $discount_codes ) ) {
+			return null;
+		}
+
+		$current_time = current_time( 'timestamp' ); // phpcs:disable WordPress.DateTime.CurrentTimeTimestamp
```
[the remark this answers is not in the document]
iamdharmesh: I think this makes sense.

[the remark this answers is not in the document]
iamdharmesh: Seems fixed in https://github.com/woocommerce/woocommerce-square/pull/440/changes/5c418df72a2a1696fc15d8ffebfa9a6ac0413523

includes/Gateway/API/Requests/Orders.php
faisal-alvi: @iamdharmesh thanks, this has been applied.

includes/Gateway/API/Requests/Orders.php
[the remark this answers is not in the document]
faisal-alvi: This has been applied to the PR.

iamdharmesh (approved):
Thanks for the changes @faisal-alvi. Code looks good now. I wasn’t able to test it end-to-end because it requires the production environment and an actual card.

@qasumitbagthariya, please perform in-depth QA on this, and let @faisal-alvi or me know if you have any questions.

qasumitbagthariya (commented):
**QA Note:**
---

I discussed all the points mentioned below with @faisal-alvi , and we reviewed each case in detail. Posting this here to keep everyone informed that these scenarios have already been aligned upon.

When multiple tax rates are applied to an order, the tax amount displays as $0.19 in the WordPress dashboard and the Square dashboard it shows the $0.17.

there appears to be inconsistency in how the values are calculated internally when multiple taxes are involved.

Observed Behavior:
- WordPress dashboard tax amount: $0.19
- Square dashboard tax amount: $0.17
- Final order total matches on both WordPress and Square.
- Tax setting is: inclusive of tax

image

image

image



---

**Case:2** 

When multiple tax rates and multiple coupons are applied to an order, a variation in the final discount amount is observed between WooCommerce and the Square dashboard.

Observed Behavior:
- In WooCommerce, the final total displays as $95.81.
- In the Square dashboard, the final total displays as $95.94.
- Final order total matches on both WordPress and Square.

There is a difference of $0.13 between the two systems.

Tax setting is: inclusive of tax

image

image


---

**Case 3:**

If the admin manually removes the coupon from the order or creates/edits the order directly from the WordPress dashboard, it does not work as expected.


image

faisal-alvi:
@qasumitbagthariya Thanks for the feedback. As discussed [internally](https://fueled.slack.com/archives/C01UCQGPBDJ/p1772472582524019?thread_ts=1772446955.167209&cid=C01UCQGPBDJ), cases 1 & 2 are related to rounding differences between the two platforms (WooCommerce and Square). Each platform rounds values based on its own settings, so there’s no actual bug here — this is handled via adjustment and is expected behavior.

Case 3 is an edge case where a merchant would need to apply a coupon ‘manually’ from the backend on an order that has already been placed. This is quite rare and not something we can address in Phase 1, so we’d flag it as a future enhancement only if there’s a specific merchant requirement for it.

qasumitbagthariya:
QA  Update:  ✅ 
---

@faisal-alvi 

Thanks for the clarification and detailed explanation. That makes sense. Since these behaviors are expected and the edge case can be considered for a future enhancement if needed, I’m approving this PR.


**Testing Environment** 

<Details>

- WordPress: 6.9.1
- WooCommerce - 10.5.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.2 
- Branch: do/SQUARE-219-discount-codes

</Details>

qasumitbagthariya (approved):

qasumitbagthariya:
-------

Status - Working expected with Plugin Archive/Zip file same as fix specific branch.

**Testing Environment** 

<Details>

- WordPress: 6.9.1
- WooCommerce - 10.5.3
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>

Next Step- Ready to Merge 🚀

PR #408, opened 2025-09-09
Add Support for Multiple Categories Sync

faisal-alvi:
### All Submissions:

* [x] Does your code follow the [WooCommerce Sniffs](https://github.com/woocommerce/woocommerce-sniffs/) variant of WordPress coding standards?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully run tests with your changes locally?
* [ ] Will this change require new documentation or changes to existing documentation?

---

### Changes proposed in this Pull Request:

This PR implements support for multiple categories in WooCommerce-Square synchronization, addressing the limitation where only one category was being synced despite both Square and WooCommerce supporting multiple categories.

- **Woo_SOR.php**: Updated `update_catalog_item()` method to sync all mapped categories instead of just the first one
- **Product_Import.php**: Enhanced `extract_product_data()` method to import all categories from Square items

Closes https://linear.app/a8c/issue/SQUARE-110/syncs-only-one-category-overwriting-multiple-even-though-both-square.

### Steps to test the changes in this Pull Request:

1. **Test WooCommerce → Square Sync:**
   - Sync the product from WooCommerce to Square
   - Verify in Square dashboard that the product has all assigned categories
   - Confirm the first category is set as the reporting category

2. **Test Square → WooCommerce Sync:**
   - Create/update a product in Square with multiple categories
   - Sync from Square to WooCommerce
   - Verify the WooCommerce product has all categories assigned correctly

3. **Test Edge Cases:**
   - Products with no categories (should sync without errors)
   - Existing single-category products (should continue working as before)

### Changelog entry

> Add - Support for multiple categories in WooCommerce-Square product synchronization.

PR #451, opened 2026-01-21
Fix Categories were being stripped from products during manual sync.

iamdharmesh:
### All Submissions:

---

In PR #408, we added support for syncing multiple categories. However, we missed adding this support to manual synchronization. This PR fixes an issue in manual and interval synchronization where categories were being stripped from products synced from Square.

> [!NOTE]
> While working on this, I noticed that synced categories do not maintain the parent–child relationship during synchronization. I have reported a separate [issue](https://linear.app/a8c/issue/SQUARE-236/synced-categories-do-not-preserve-parent-child-hierarchy) to explore implementation options and potentially address it.

Closes https://linear.app/a8c/issue/SQUARE-231/categories-get-stripped-off-from-products-that-are-synced-from-square

1. **Square SOR:**

   * Create a product in Square with multiple categories.
   * Sync from Square to WooCommerce (via import).
   * Verify that the WooCommerce product has all categories assigned correctly.
   * Update the product on the Square side (e.g., update the product title).
   * Wait for the interval polling to trigger (or trigger it manually).
   * Verify that the WooCommerce product still has all categories assigned correctly and that none are removed.
   * Trigger a manual sync from the settings and verify that the WooCommerce product still has all categories assigned correctly and that none are removed.

2. **WooCommerce SOR:**

   * Create a product in WooCommerce with multiple categories and sync it with Square.
   * Verify that the Square product has all categories assigned correctly.
   * Update the product on the WooCommerce side (e.g., update the product title).
   * Verify that the Square product still has all categories assigned correctly and that none are removed.
   * Trigger a manual sync from the settings and verify that the Square product still has all categories assigned correctly and that none are removed.

### DIFF
--- a/includes/Handlers/Category.php
+++ b/includes/Handlers/Category.php
@@ -271,7 +271,8 @@ class Category {
 	 * @return array
 	 */
 	public static function get_category_ids_from_catalog_item( $catalog_item ) {
-		$category_ids = array();
+		$category_ids         = array();
+		$missing_category_ids = array();
 
 		if ( empty( $catalog_item ) ) {
 			return $category_ids;
--- a/includes/Handlers/Category.php
+++ b/includes/Handlers/Category.php
@@ -267,6 +267,8 @@ class Category {
 	/**
 	 * Get the WooCommerce category IDs from a catalog item.
 	 *
+	 * @since x.x.x
+	 *
 	 * @param \Square\Models\CatalogItem $catalog_item the catalog item object
 	 * @return array
 	 */
@@ -274,7 +276,7 @@ class Category {
 		$category_ids         = array();
 		$missing_category_ids = array();
 
-		if ( empty( $catalog_item ) ) {
+		if ( empty( $catalog_item ) || ! $catalog_item instanceof \Square\Models\CatalogItem ) {
 			return $category_ids;
 		}
 
--- a/includes/Handlers/Category.php
+++ b/includes/Handlers/Category.php
@@ -263,4 +263,53 @@ class Category {
 
 		return $catalog_category_id;
 	}
+
+	/**
+	 * Get the WooCommerce category IDs from a catalog item.
+	 *
+	 * @param \Square\Models\CatalogItem $catalog_item the catalog item object
+	 * @return array
+	 */
+	public static function get_category_ids_from_catalog_item( $catalog_item ) {
+		$category_ids = array();
+
+		if ( empty( $catalog_item ) ) {
+			return $category_ids;
+		}
+
+		if ( $catalog_item->getCategories() && is_array( $catalog_item->getCategories() ) ) {
+			foreach ( $catalog_item->getCategories() as $category ) {
+				if ( $category instanceof \Square\Models\CatalogObjectCategory ) {
+					$category_id = self::get_category_id_by_square_id( $category->getId() );
+					if ( $category_id ) {
+						$category_ids[] = $category_id;
+					} else {
+						$missing_category_ids[] = $category->getId();
+					}
+				}
+			}
+		}
+
+		// Fetch and import missing categories.
+		if ( ! empty( $missing_category_ids ) ) {
+			try {
+				$response = wc_square()->get_api()->batch_retrieve_catalog_objects( $missing_category_ids );
+				if ( $response->get_data() instanceof \Square\Models\BatchRetrieveCatalogObjectsResponse ) {
+					$missing_categories = $response->get_data()->getObjects();
+					if ( $missing_categories && is_array( $missing_categories ) ) {
+						foreach ( $missing_categories as $missing_category ) {
+							$imported_category_id = self::import_or_update( $missing_category );
+							if ( $imported_category_id ) {
+								$category_ids[] = $imported_category_id;
+							}
+						}
+					}
+				}
+			} catch ( \Exception $e ) {
+				wc_square()->log( 'Error fetching missing categories for product ' . $catalog_item->getName() . ': ' . $e->getMessage() );
+			}
+		}
+
+		return array_unique( $category_ids );
+	}
 }
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -230,33 +230,11 @@ class Product {
 			$product->set_description( $product_description );
 		}
 
-		$square_category_id = Category::get_square_category_id( $catalog_item );
-		$category_id        = Category::get_category_id_by_square_id( $square_category_id );
+		$category_ids = Category::get_category_ids_from_catalog_item( $catalog_item );
 
-		if ( $category_id ) {
-			wp_set_object_terms( $product->get_id(), intval( $category_id ), 'product_cat' );
-		} else {
-			$message = sprintf(
-				/* translators: Placeholder: %s category ID */
-				__( 'Square category with id (%s) was not imported to your Store. Please run Import Products from Square settings.', 'woocommerce-square' ),
-				$square_category_id
-			);
-
-			$records = Records::get_records();
-			foreach ( $records as $record ) {
-				if ( $record->get_message() === $message ) {
-					$is_recorded = true;
-				}
-			}
-
-			if ( ! isset( $is_recorded ) ) {
-				Records::set_record(
-					array(
-						'type'    => 'alert',
-						'message' => $message,
-					)
-				);
-			}
+		if ( ! empty( $category_ids ) ) {
+			$term_ids = array_unique( array_map( 'intval', $category_ids ) );
+			wp_set_object_terms( $product->get_id(), $term_ids, 'product_cat' );
 		}
 
 		if ( $catalog_id ) {
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -230,13 +230,6 @@ class Product {
 			$product->set_description( $product_description );
 		}
 
-		$category_ids = Category::get_category_ids_from_catalog_item( $catalog_item );
-
-		if ( ! empty( $category_ids ) ) {
-			$term_ids = array_unique( array_map( 'intval', $category_ids ) );
-			wp_set_object_terms( $product->get_id(), $term_ids, 'product_cat' );
-		}
-
 		if ( $catalog_id ) {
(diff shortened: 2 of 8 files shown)


includes/Handlers/Category.php
```diff
@@ -263,4 +263,53 @@ public static function get_square_category_id( $catalog_item ) {
...
+				if ( $category instanceof \Square\Models\CatalogObjectCategory ) {
+					$category_id = self::get_category_id_by_square_id( $category->getId() );
+					if ( $category_id ) {
+						$category_ids[] = $category_id;
+					} else {
+						$missing_category_ids[] = $category->getId();
+					}
+				}
+			}
+		}
```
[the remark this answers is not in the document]
iamdharmesh: If `getCategories` is empty then there is not reporting category.

tests/e2e/specs/c4.product-import.spec.js
```diff
@@ -70,7 +79,8 @@ test( 'Import Cap from Square @sync', async ( { page, baseURL } ) => {
 	await expect( await page.locator( '.entry-summary .sku_wrapper' ) ).toHaveText( 'SKU: cap-regular' );
 	await expect( await page.getByText( 'This is a very good cap, no cap.' ) ).toBeVisible();
 	await expect( await page.getByText( '53 in stock' ) ).toBeVisible();
-	await expect( await page.getByText( 'Category: Hats' ) ).toBeVisible();
+	await expect( page.getByText( 'Hats' ) ).toBeVisible();
+	await expect( page.getByText( 'Hats2' ) ).toBeVisible();
```
iamdharmesh:
```suggestion
	await expect( await page.getByText( 'Categories: Hats, Hats2' ) ).toBeVisible();
```

includes/Handlers/Category.php
```diff
@@ -263,4 +263,54 @@ public static function get_square_category_id( $catalog_item ) {
 
 		return $catalog_category_id;
 	}
+
+	/**
+	 * Get the WooCommerce category IDs from a catalog item.
+	 *
```
faisal-alvi, line 269: Suggestion: Add `@since` to the docblock to let our future versions know when this API method was introduced.

includes/Handlers/Category.php
```diff
@@ -263,4 +263,54 @@ public static function get_square_category_id( $catalog_item ) {
...
+	 */
+	public static function get_category_ids_from_catalog_item( $catalog_item ) {
+		$category_ids         = array();
+		$missing_category_ids = array();
+
+		if ( empty( $catalog_item ) ) {
+			return $category_ids;
+		}
+
+		if ( $catalog_item->getCategories() && is_array( $catalog_item->getCategories() ) ) {
```
faisal-alvi, line 283: I know this part is not new and just moved here, but just a nit, should we consider validating that `$catalog_item` has a `getCategories()` method (e.g. accept only `\Square\Models\CatalogItem` in the beginning) to avoid runtime errors if a wrong type is passed?

includes/Handlers/Category.php
```diff
@@ -263,4 +263,54 @@ public static function get_square_category_id( $catalog_item ) {
...
+				}
+			}
+		}
+
+		// Fetch and import missing categories.
+		if ( ! empty( $missing_category_ids ) ) {
+			try {
+				$response = wc_square()->get_api()->batch_retrieve_catalog_objects( $missing_category_ids );
+				if ( $response->get_data() instanceof \Square\Models\BatchRetrieveCatalogObjectsResponse ) {
+					$missing_categories = $response->get_data()->getObjects();
```
faisal-alvi, line 301: if `getObjects()` returns an empty array (e.g. deleted/invalid IDs), we only log on exception. Should we consider logging when some IDs were requested but none returned, for “category still missing” cases?

includes/Handlers/Product.php
```diff
@@ -230,33 +230,11 @@ public static function update_from_square( \WC_Product $product, \Square\Models\
...
-				Records::set_record(
-					array(
-						'type'    => 'alert',
-						'message' => $message,
-					)
-				);
-			}
+		if ( ! empty( $category_ids ) ) {
+			$term_ids = array_unique( array_map( 'intval', $category_ids ) );
+			wp_set_object_terms( $product->get_id(), $term_ids, 'product_cat' );
```
faisal-alvi: When `$category_ids` is empty (Square has no categories for this product), we do not call `wp_set_object_terms()`. So existing Woo categories on the product are left unchanged. Is that intentional? If the contract is “Woo categories = Square categories,” we should clear when empty, e.g. `wp_set_object_terms( $product->get_id(), array(), 'product_cat' );` in an else branch.

tests/e2e/specs/c2.woo-sor.spec.js
```diff
@@ -78,15 +79,21 @@ test( 'iPhone Pro pushed to Square @sync', async ( { page } ) => {
...
 			categoryName = categories.objects
 				.filter((cat) => cat.id === category)
 				.map((cat) => cat.category_data.name)[0];
+			categoriesNames = categories.objects
+				.filter((cat) => categoriesIds.includes(cat.id))
+				.map((cat) => cat.category_data.name);
 		}
 	}
 	expect(categoryName).toEqual('Mobiles');
+	expect(categoriesNames).toContain('Mobiles');
```
faisal-alvi, line 95: If `category` is falsy, `categoriesNames` stays `[]` and `expect(categoriesNames).toContain('Mobiles')` will fail. Should we consider asserting that `categoriesIds.length >= 1` (or that `category` is set) so the failure message is clearer when the push doesn’t return categories?

tests/e2e/specs/d8.square-sor-multiple-variations.spec.js
```diff
@@ -104,25 +104,25 @@ test( '[Square SOR] Import multiple variations products from Square @sync', asyn
 	await page.locator( 'li.variations_tab a' ).click();
 	await expect(
 		page.locator( 'select[name="attribute_pa_color[0]"]' )
-	).toHaveValue( 'blue' );
+	).toHaveValue( 'red' );
```
faisal-alvi, line 107: The only changes in this file are swapping expected colors (e.g. 'red' ↔ 'blue') for the same variation indices. This is unrelated to the PR. Why do we have to change these? Were the tests failing due to changes in this PR?

faisal-alvi (changes requested):
@iamdharmesh thanks for working on this!

When Square has no categories, we should explicitly set product categories to empty so Woo matches Square, or document that we intentionally leave existing categories. I have also added a few suggestions. Please have a look.

tests/e2e/specs/d8.square-sor-multiple-variations.spec.js
iamdharmesh: That is because the current Square account configured in E2E has many categories, which prevented us from comparing the created categories since they were not returned in the initial results, and we do not have access to that account to delete/clean category. Therefore, I updated the Square account credentials which are used in the E2E, to use an account with less data and proper access. However, in this new account, the order of attributes has changed, so I had to update it accordingly.

tests/e2e/specs/c2.woo-sor.spec.js
iamdharmesh: If the category has a falsy value, `expect(categoryName).toEqual('Mobiles');` will fail and display both the expected and actual values in the failure log, which I think is sufficient to identify that the category was not pushed or set on Square. However, I’m open to adding that as well if you have a strong preference.

includes/Handlers/Category.php
iamdharmesh: Added in https://github.com/woocommerce/woocommerce-square/pull/451/changes/1ccb8deb2168745e8526a560eaeed2ebbdd790ca

includes/Handlers/Category.php
iamdharmesh: Added in https://github.com/woocommerce/woocommerce-square/pull/451/changes/1ccb8deb2168745e8526a560eaeed2ebbdd790ca

includes/Handlers/Category.php
iamdharmesh: We log exception if we didn't get a proper response, but not logging for the empty array? Could you please provide more information on what we want to do here? Thanks

includes/Handlers/Product.php
iamdharmesh: I have removed this as this is no longer needed.

iamdharmesh:
@faisal-alvi this is ready for your review.

includes/Handlers/Category.php
faisal-alvi: Since the IDs we're requesting were just returned from the recent catalog item, we're unlikely to receive an empty array of get objects at this point. Resolving this thread.

faisal-alvi (approved):

ankitguptaindia:
🧪  QA/Test Status: Approved ✅ 
------------


Tested the PR changes for both Square SOR and WooCommerce SOR flows.

Verified that products with multiple categories retain all assigned categories correctly after:

* initial sync/import
* product updates
* interval polling sync
* manual sync execution

Confirmed that no categories were removed unexpectedly during sync operations on either Square or WooCommerce side. PR works as expected.





### Functional Demo /  Screencast: 


**When Square SOR:**


| Product in Woo | Product in Squre |
|--------|--------|
| image | image |








**When WooCommerce SOR:**

| Product in Woo | Product in Squre |
|--------|--------|
| image | image |





### 🐛 Issues Found: 
None


### 🖥 Testing Environment: 

<Details>

- WordPress: 7.0
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 148.0.0.0
- OS: macOS

</Details>


### 🔁 Next Step: 
Ready to merge

ankitguptaindia (approved):

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 7.0
- WooCommerce - 10.8.0
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>

🔁 Next Step- Ready to Merge 🚀

PR #478, opened 2026-03-31
Skip Error Throws for Stale Square Catalog Errors

faisal-alvi:
### All Submissions:

---

- Introduce a dedicated `COUPON_REDEMPTION_ERROR_CODE` constant on the `Gateway` class to tag coupon/redemption-related exceptions that should fail checkout.
- Update the coupon redemption error paths in `do_transaction()` to throw exceptions with this explicit error code when Square discount redemptions fail (e.g. redemption limit, expired code, missing redemption ID).
- Narrow the outer `do_transaction()` catch block to only re-throw exceptions when their code matches `COUPON_REDEMPTION_ERROR_CODE`, logging and swallowing other `create_order()`/catalog-related exceptions so checkout can proceed as in pre-5.3.0 behavior.

Closes https://linear.app/a8c/issue/SQUARE-276/regression-v530-throw-dollarexception-in-do-transaction-breaks.

Closes https://linear.app/a8c/issue/SQUARE-277/regression-in-530-checkout-fails-with-value-too-low-base-price-money

For SQUARE-277, I was able to reproduce the issue using a paid plugin. While investigating a fix, I discovered that these third-party plugins add discounts as line items rather than proper “discount” items - which doesn’t align with the Square checkout flow. A potential fix would be to convert those line items into Discounts, but that carries a high risk of regression.

After digging deeper, I found the root cause: we started surfacing (i.e. `throw`-ing) these errors with the Coupon Code release, and the errors themselves are legitimate. However, this PR - which filters the thrown exceptions - also resolves this issue. I tested against it and confirmed SQUARE-277 is fixed: the line items are rejected by Square, but the order no longer fails, which matches the behavior we had prior to the Coupon Code release. So I’d recommend keeping that approach as-is.

1. Configure a store where a Square coupon/discount code is valid and successfully redeemable, then place an order using that code and confirm checkout still succeeds and the coupon is applied correctly.
2. Configure a store with a Square coupon/discount that cannot be redeemed (e.g. exhausted redemption limit or expired) and confirm checkout fails with the expected coupon error message (and not a generic payment decline).
3. Configure a product with stale Square catalog IDs (e.g. via disconnect/reconnect or Square-side deletion), attempt checkout with and without coupons, and confirm checkout now completes successfully while the catalog NOT_FOUND error is logged instead of being surfaced as a payment decline.
4. Also confirm the issue SQUARE-277 is also fixed.

### Changelog entry

> Fix - Prevent stale Square catalog IDs from breaking checkout while still surfacing genuine coupon redemption failures as checkout errors.

### DIFF
--- a/includes/Gateway.php
+++ b/includes/Gateway.php
@@ -53,7 +53,7 @@ class Gateway extends Payment_Gateway_Direct {
 	/**
 	 * Error code used to indicate coupon/redemption failures that should fail checkout.
 	 *
-	 * @since 5.3.1
+	 * @since 5.3.2
 	 */
 	const COUPON_REDEMPTION_ERROR_CODE = 9001;
 
--- a/includes/Gateway.php
+++ b/includes/Gateway.php
@@ -50,6 +50,13 @@ use WooCommerce\Square\Utilities\Performance_Logger;
  */
 class Gateway extends Payment_Gateway_Direct {
 
+	/**
+	 * Error code used to indicate coupon/redemption failures that should fail checkout.
+	 *
+	 * @since 5.3.1
+	 */
+	const COUPON_REDEMPTION_ERROR_CODE = 9001;
+
 
 	/** @var Gateway\API API base instance */
 	private $api;
@@ -511,7 +518,8 @@ class Gateway extends Payment_Gateway_Direct {
 										/* translators: %s: error message from Square */
 										__( 'The coupon could not be applied (e.g. redemption limit reached or code expired). Please remove it and try again, or use a different payment method. Error: %s', 'woocommerce-square' ),
 										$error_message
-									)
+									),
+									self::COUPON_REDEMPTION_ERROR_CODE
 								);
 							}
 
@@ -520,7 +528,8 @@ class Gateway extends Payment_Gateway_Direct {
 									$this->get_plugin()->log( sprintf( 'Square: Create redemption returned no id for discount code %1$s on order #%2$s', $square_discount_code_id, $order->get_id() ), $this->get_id() );
 								}
 								throw new \Exception(
-									__( 'The coupon could not be applied. Please remove it and try again, or use a different payment method.', 'woocommerce-square' )
+									__( 'The coupon could not be applied. Please remove it and try again, or use a different payment method.', 'woocommerce-square' ),
+									self::COUPON_REDEMPTION_ERROR_CODE
 								);
 							}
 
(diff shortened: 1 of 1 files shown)


includes/Gateway.php
```diff
@@ -50,6 +50,13 @@
  */
 class Gateway extends Payment_Gateway_Direct {
 
+	/**
+	 * Error code used to indicate coupon/redemption failures that should fail checkout.
+	 *
+	 * @since 5.3.2
```
iamdharmesh, line 56:
```suggestion
	 * @since x.x.x
```

iamdharmesh (approved):
Thanks @faisal-alvi. Looks good to me.

qasumitbagthariya (approved):
### QA Update  ✅ 
-------------

I have verified this PR in the `SQUARE-276` branch, which has been fixed and is functioning as intended.

- Verified checkout with a valid Square coupon/discount code, the coupon was applied correctly and checkout completed successfully.
- Verified behavior with an invalid/non-redeemable Square coupon (expired/exhausted), checkout failed with the appropriate coupon-specific error message (no generic payment decline shown).
- Tested with products having stale Square catalog IDs (via disconnect/reconnect scenario), checkout completed successfully both with and without coupons, and the catalog NOT_FOUND error was logged without impacting the user-facing payment flow.
- Confirmed that the issue SQUARE-277 is resolved.

**Before fix**

image


**After fix**

image

image


**exhausted redemption limit**
image


Fix [Square#277](https://linear.app/a8c/issue/SQUARE-277/regression-in-530-checkout-fails-with-value-too-low-base-price-money)

https://github.com/user-attachments/assets/5fd50922-9479-402e-9bbb-35a474dd2bc6


image


**Testing Environment** 

<Details>

- WordPress: 6.9.4
- WooCommerce - 10.6.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.2 
- Branch: SQUARE-276

</Details>


Testing Document status:

- [ ] Yes

qasumitbagthariya:
-------

**Testing Environment** 

<Details>

- WordPress: 6.9.4
- WooCommerce - 10.6.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome 
- OS: macOS 15.2 
- Branch: smoke-testing

</Details>

Next Step- Ready to Merge 🚀

Issue #232, opened 2024-11-02
Duplicates on upload/sync


grig-23:
### Issue

In 8942863-zen, On upload/sync, we get duplicate products created, and these duplicates correspond to errors such as:

`Product #3471 is excluded from sync as the product type "variation" is unsupported.`

This results in a stub product in Square where 3471 and it's SKU is showing up in what I would call an orphaned state, , as well as a properly synced product, which contains all the variations, #3471 and #3472 etc.

429 rate limiting causes duplicates #108 but we have reset the batch sizes, and rate limiting errors are no longer occurring.

SOR - WooCommerce
Square has a clean DB

The merchant tried cleaning products from Square and re-syncing, only to see more duplicates.

p1730380630885499-slack-C01TYRCJ1QD

nicdwilson:
I have imported the customers products into a test site, and can replicate the validation error.

`Product #3471 is excluded from sync as the product type "variation" is unsupported.`

However, my products are not syncing at all to the sandbox: I have no items in the library, WooCommerce is SOR.

I did find some reports that look very similar on the Square forums [such as here](https://www.sellercommunity.com/t5/Square-Online/Product-type-quot-variation-quot-is-unsupported-error-when/td-p/721947)

qasumitbagthariya:
QA update:
---

I attempted to reproduce the issue on my end but was unable to do so. Here's what I tested:

- Platforms: SOR WooCommerce and Square
- Product: Variable  
- WordPress Version: 6.7
- PHP Version: 8.0.30
- Web Server: Nginx 1.20.2
- Branch: trunk

iamdharmesh:
Hi @grig-23,

Thank you for reporting the issue, and @nicdwilson, thanks for providing the additional details.

I’ve tested this across multiple scenarios but haven’t been able to reproduce the issue. However, I was able to replicate the error `Product #XXXX is excluded from sync as the product type "variation" is unsupported` by making manual changes to the database. Based on my findings, though, this doesn’t appear to be the root cause of the duplication issue.

I’m not sure what we might be missing when attempting to replicate the issue, but I wonder if it could be related to specific product data? 🤔 

> I have imported the customers products into a test site, and can replicate the validation error.

@nicdwilson, could you share the exported data of the product causing the issue so I can replicate it on my end and investigate further? Any additional details would also be appreciated.

Thank you!

nicdwilson:
[...]

Issue #108, opened 2024-04-23
Duplicates in the Square catalog when hitting a rate limiting error


a-danae:
### Describe the bug

During catalog syncing, duplicates are created in the Square catalog when hitting a rate limiting error on Square's end.

Square's support insights on the duplicates:

> It does look like there is something off with these requests - the rate limiting error coming from us is not entirely correct, as some portion of the batch upsert request is successfully completing. This is why duplicates are being created; following the RATE_LIMITED error WooCommerce believes the entire request was not successful and retries it but some of items were already successfully created.

### To reproduce
To be replicated on our end, but this is what the merchant is doing is having a big set of products sync with Square. Enough to hit the rate limit.

### Expected behavior

If a rate limiting error is reached, there must not be duplicates when resuming the sync.

### Additional details

From Nic Wilson:

> The rate limit belongs to the Square account, which can be doing things other than interacting with WooCommerce. In this case, the account is also syncing the catalogue to Square Online. This means the limit is not something we can manage, so is not really our issue. Square acknowledge that.
However, the pickup and retry after the limit is applied is creating duplicate objects. This appears to be because the entire batch is rerun, even though half the batch may have been successful before failure. My guess is reusing the idempotency key helps prevent duplication of catalogue items.

Related idempotency key issue https://github.com/woocommerce/woocommerce-square/issues/106

8051230-zd-a8c
p1713482527612829-slack-C7U3Y3VMY

nicdwilson:
Additional input from Square regarding 8051230-zd-a8c

```
Essentially the rate limiting error is not real in this case; one of the batches in the request is being rate limited by one of the prior batches not releasing the ‘catalog lock’ which is put into place for concurrency reasons.

This endpoint is also fairly problematic in general. Different batches in the request can succeed or fail independently, but for non 200 response scenarios the error handling is not sufficient for your client to determine which batches failed versus succeeded.

The team is currently considering the longer term fixes to this endpoint. But in the interim to get this merchant unblocked we recommended:

- when receiving a 429 or 5XX API response from Square, the integration should retry with the same idempotency_key used in the original request. This will prevent retries from creating duplicate items if some of the items were actually successfully created in the original request.

- adjust the integration to only upsert a single batch in each request, since the error handling for failure cases with multiple batches is not being done well on our side currently

```

Note related issue #106

nicdwilson:
[...]

PR #483, opened 2026-04-08
fix: add SKU-based existence check before creating new Square catalog items

slash1andy:
## Summary

When a sync job times out during `search_matched_products`, the retry can create duplicate Square catalog items. This happens because `upsert_new_products()` creates new items without checking if a product with the same SKU already exists in the Square catalog.

**Root cause:** Steps 8 (`search_matched_products`) and 9 (`upsert_new_products`) use different idempotency keys. If step 8 partially completes and times out after creating some Square items, step 9 sees those products as "unprocessed" and creates duplicates.

**Observed impact:** One merchant with 1,100+ products ended up with three copies of the same product in Square, all with identical SKUs.

## Fix

Adds a pre-creation guard in `upsert_new_products()` that:

1. For each product about to be upserted, searches the Square catalog for items with a matching SKU
2. If an exact SKU match is found, links the existing Square item to the WooCommerce product (`set_square_item_id`) instead of creating a duplicate
3. Logs the match for visibility
4. If the lookup fails (network error, rate limit), logs the failure and proceeds with the upsert — the worst case is the existing duplicate behavior, not a new crash

This is the same defensive pattern recommended in the related issues [#108](https://github.com/woocommerce/woocommerce-square/issues/108) and [#232](https://github.com/woocommerce/woocommerce-square/issues/232).

## Files changed

- `includes/Sync/Manual_Synchronization.php` — `upsert_new_products()` method

## Performance note

This adds one `SearchCatalogObjects` API call per product in the upsert batch. For large batches this could be significant. A future optimization could batch the SKU lookups into a single search, but the current approach prioritizes correctness and safety over speed for this critical data-integrity fix.

## Test plan

- [ ] Normal sync with no prior timeout — verify no regression, products create as expected
- [ ] Simulate a timeout during `search_matched_products` (e.g., reduce AS timeout), then re-trigger sync — verify no duplicate items created
- [ ] Product with no SKU — verify it passes through to upsert without the lookup
- [ ] Product with a SKU that already exists in Square — verify it links instead of duplicating
- [ ] API lookup failure (e.g., rate limit) — verify sync continues and logs the failure

Fixes SQUARE-290.

### DIFF
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -27,6 +27,8 @@ use Square\Models\BatchRetrieveInventoryCountsResponse;
 use Square\Models\BatchUpsertCatalogObjectsResponse;
 use Square\Models\BatchRetrieveCatalogObjectsResponse;
 use Square\Models\CatalogObject;
+use Square\Models\CatalogQuery;
+use Square\Models\CatalogQueryText;
 use Square\Models\SearchCatalogObjectsResponse;
 use Square\Models\CatalogInfoResponse;
 use Square\ApiHelper;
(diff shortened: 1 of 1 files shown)


includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
 			$this->set_attr( 'upsert_new_product_ids', array() );
 		}
 
+		// Before creating new Square items, check if products already exist in Square
+		// by SKU. This prevents duplicate catalog items when a previous sync timed out
+		// during search_matched_products and partially created items. See SQUARE-290.
+		$skipped_product_ids = array();
+
+		foreach ( $product_ids as $key => $product_id ) {
```
faisal-alvi: Rather than adding the code inline, it'd be cleaner to move it into its own method. I actually did something similar in another PR, and actually for the same purpose #471.

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+					array(
+						'object_types' => array( 'ITEM' ),
+						'query'        => array(
+							'text_query' => array(
+								'keywords' => array( $sku ),
+							),
+						),
+						'limit'        => 10,
+					)
+				);
```
faisal-alvi: The `set_search_catalog_objects_data()` passes `$args['query']` straight into `SearchCatalogObjectsRequest::setQuery()`. The SDK expects a `Square\Models\CatalogQuery` instance, not a nested PHP array. With the current Square PHP SDK this should be a `TypeError` (or equivalent) at runtime. Did this work in your testing without an error?

I think we need a real SDK objects, e.g. `CatalogQuery` + `CatalogQueryText` (or preferably `CatalogQueryExact` on attribute sku for variation-level matching). See the following code for reference.

https://github.com/woocommerce/woocommerce-square/blob/92bfdec6e798d235efd89f790c491eb2331f236d/includes/Handlers/Product.php#L1190

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+			try {
+				$response = wc_square()->get_api()->search_catalog_objects(
+					array(
+						'object_types' => array( 'ITEM' ),
+						'query'        => array(
+							'text_query' => array(
+								'keywords' => array( $sku ),
+							),
+						),
+						'limit'        => 10,
```
faisal-alvi: Capping at 10 objects can miss the item that actually has the matching variation if text ranking puts it outside the first page, so we can still create duplicates. Should we consider a higher limit, pagination, or an exact SKU query that returns the right rows directly?

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+		$skipped_product_ids = array();
+
+		foreach ( $product_ids as $key => $product_id ) {
+			$product = wc_get_product( $product_id );
+
+			if ( ! $product ) {
+				continue;
+			}
+
+			$sku = $product->get_sku();
```
faisal-alvi: Here the `$product->get_sku()` is used only on the parent. Many variable products have an empty parent SKU; SKUs live on variations. Those products continue with no check, variation SKUs should be included in the guard or those products will bypass it entirely.

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+				// If the lookup fails, proceed with the upsert — the worst case
+				// is a duplicate that can be cleaned up, not a crashed sync.
+				wc_square()->log( sprintf(
+					'SQUARE-290 guard: SKU lookup failed for product #%d (SKU: %s): %s. Proceeding with upsert.',
+					$product_id,
+					$sku,
+					$e->getMessage()
+				) );
+			}
+		}
```
faisal-alvi: On API failure we are still allowing process anyway. Doesn't it keep the duplicate window open whenever the lookup fails? Maybe worth calling out docs or log so we/merchants know duplicates are still possible when the API errors or times out.

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+							// Found an existing Square item with the same SKU — link it
+							// instead of creating a duplicate.
+							wc_square()->log( sprintf(
+								'SQUARE-290 guard: Product #%d (SKU: %s) already exists in Square as %s. Linking instead of creating duplicate.',
+								$product_id,
+								$sku,
+								$existing_object->getId()
+							) );
+
+							Product::set_square_item_id( $product_id, $existing_object->getId() );
```
faisal-alvi: We are setting only `square_item_id` here, this skips the usual `Woo_SOR::update_product` / `update_variation` flow (versions, variation IDs, image meta, etc.). Should we prefer aligning with the existing “matched product” path: link fully, then `upsert_catalog_objects(..., false)` (or equivalent) to avoid the risk of inconsistent meta?

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +745,94 @@ protected function upsert_new_products() {
...
+							$skipped_product_ids[] = $product_id;
+
+							// Break out of both foreach loops for this product.
+							break 2;
```
faisal-alvi: If there have multiple Square items share the same SKU. Should we detect more than one distinct catalog item (or variation) that matches and skip with a warning instead of picking arbitrarily?

faisal-alvi (changes requested):
@slash1andy Thanks for the PR!

I've left a few inline comments below. Also worth noting - the PHP coding standards checks and some E2E tests are currently failing in the pipeline. Could you take a look at those as well?

faisal-alvi:
Requesting review from @iamdharmesh - since another PR (#471) also targets the same file and uses a similar SKU-based lookup approach, it made sense to keep the review consistent.

faisal-alvi:
@slash1andy while Dharmesh's review is pending, could you take a look at the PHPCS failures in the pipeline and address them?

includes/Sync/Manual_Synchronization.php
```diff
@@ -717,6 +719,194 @@ protected function search_matched_products() {
...
 
 
+	/**
+	 * Searches the Square catalog for an existing item that matches a WooCommerce product by SKU.
+	 *
+	 * Collects SKUs from both the parent product and its variations, then queries the Square
+	 * catalog using a text search. Results are verified with an exact SKU comparison against
+	 * variation-level data from the returned catalog objects.
+	 *
+	 * @since 4.9.0
```
iamdharmesh, line 729: Nit:
```suggestion
	 * @since x.x.x
```
This will be replaced during release.

includes/Sync/Manual_Synchronization.php
```diff
@@ -717,6 +719,194 @@ protected function search_matched_products() {
...
+	}
+
+
+	/**
+	 * Links a WooCommerce product to an existing Square catalog object by setting item and variation IDs.
+	 *
+	 * This follows the same metadata pattern used in square_sor_sync() to ensure consistent
+	 * state across both the item and variation levels.
+	 *
+	 * @since 4.9.0
```
iamdharmesh, line 862:
```suggestion
	 * @since x.x.x
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -717,6 +719,194 @@ protected function search_matched_products() {
...
+		// Search Square catalog using product SKUs as keywords. Text search is fuzzy
+		// (prefix-based across multiple fields), so we verify with exact matching below.
+		$query = new CatalogQuery();
+		$text_query = new CatalogQueryText( $skus );
+		$query->setTextQuery( $text_query );
+
+		try {
+			$response = wc_square()->get_api()->search_catalog_objects(
+				array(
+					'object_types' => array( 'ITEM' ),
```
iamdharmesh, line 771: I think we have to query for `ITEM_VARIATION` as we are using SKUs in query.

iamdharmesh (commented):
Thanks for the PR @slash1andy. Added some notes.

> Requesting review from @iamdharmesh - since another PR (https://github.com/woocommerce/woocommerce-square/pull/471) also targets the same file and uses a similar SKU-based lookup approach, it made sense to keep the review consistent.

@faisal-alvi both PR has SKU-based lookup, but this PR use that in upsert products step while #471 is using it in push inventory. So, I am more inclined towards this.

lynnjat7:
Checking in on this one as we have a merchant who is waiting for this fix. Any thoughts on moving it forward?

vikrampm1:
Hi @lynnjat7 - We were waiting for the above feedback to be addressed so we could move it forward. However, Andrew from Woo seems to be unavailable, so our team will take it from here and get it moving. We expect to wrap it up and have it released sometime next week.

iamdharmesh:
Closing this in favour of https://github.com/woocommerce/woocommerce-square/pull/486

PR #486, opened 2026-04-29
Add SKU-based existence check before creating new Square catalog items

iamdharmesh:
### All Submissions:

---

As reported in https://linear.app/a8c/issue/SQUARE-290/timeout-during-search-matched-products-causes-duplicate-square-catalog, currently, the timeout during search_matched_products causes duplicate Square catalog items on sync retry

This PR follows the defensive fix here, basically it adds sku-based  existence check before creating new Square catalog items in the `upsert_new_products()` to prevent duplicates. (This PR follow the approach taken in https://github.com/woocommerce/woocommerce-square/pull/483)

>[!NOTE]
> The issue also mentions timeouts and other concerns as well. This PR does not address those timeouts and instead focuses on preventing existing items from being pushed again.

Closes https://linear.app/a8c/issue/SQUARE-290/timeout-during-search-matched-products-causes-duplicate-square-catalog

SOR: WooCommerce

1. There is no easy way to test this, but we can simulate the situation described in the issue with a minor code change.
2. Check out the `trunk` branch.
3. Comment out these two lines:
https://github.com/woocommerce/woocommerce-square/blob/46e9e168fb8838ae3817499e8a63c51c0ceed6e8/includes/Sync/Manual_Synchronization.php#L1722-L1723
4. Add a new product and sync it with Square.
5. Delete the Square-related metadata from the `postmeta` table for that product (`_square_*`).
6. Go to the product edit screen and click “Update.”
7. Go to the Square dashboard and verify that a duplicate product is created.
8. Check out this PR branch. Follow steps 3–6 and verify that no duplicate product is created on the Square side.

### Changelog entry

> Fix - Add SKU-based existence check before creating new Square catalog items to prevent duplicates.

### DIFF
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -23,10 +23,14 @@
 
 namespace WooCommerce\Square\Sync;
 
+use Automattic\WooCommerce\Enums\ProductType;
 use Square\Models\BatchRetrieveInventoryCountsResponse;
 use Square\Models\BatchUpsertCatalogObjectsResponse;
 use Square\Models\BatchRetrieveCatalogObjectsResponse;
 use Square\Models\CatalogObject;
+use Square\Models\CatalogObjectType;
+use Square\Models\CatalogQuery;
+use Square\Models\CatalogQuerySet;
 use Square\Models\SearchCatalogObjectsResponse;
 use Square\Models\CatalogInfoResponse;
 use Square\ApiHelper;
@@ -716,6 +720,172 @@ class Manual_Synchronization extends Stepped_Job {
 		}
 	}
 
+	/**
+	 * Links products to existing Square items by SKUs.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array $product_ids The IDs of the WooCommerce products to link.
+	 * @return array The IDs of the linked products.
+	 */
+	protected function link_products_to_existing_square_items( $product_ids ) {
+		$linked_product_ids       = array();
+		$product_skus             = array();
+		$existing_catalog_objects = array();
+
+		foreach ( $product_ids as $product_id ) {
+			$product = wc_get_product( $product_id );
+			if ( ! $product ) {
+				continue;
+			}
+
+			if ( ! empty( $product->get_sku() ) ) {
+				$product_skus[] = $product->get_sku();
+			}
+
+			if ( $product->is_type( 'variable' ) && $product->has_child() ) {
+				foreach ( $product->get_children() as $variation_id ) {
+					$variation = wc_get_product( $variation_id );
+
+					if ( $variation instanceof \WC_Product && ! empty( $variation->get_sku() ) ) {
+						$product_skus[] = $variation->get_sku();
+					}
+				}
+			}
+		}
+
+		// Remove duplicates if any.
+		$product_skus = array_unique( $product_skus );
+
+		if ( empty( $product_skus ) ) {
+			return $linked_product_ids;
+		}
+
+		// Set query has limit of 250 items.
+		foreach ( array_chunk( $product_skus, 250 ) as $batched_skus ) {
+			$existing_catalog_objects = array_merge( $existing_catalog_objects, $this->find_existing_square_items_by_skus( $batched_skus ) );
+		}
+
+		// If no existing catalog objects found, return.
+		if ( empty( $existing_catalog_objects ) ) {
+			return $linked_product_ids;
+		}
+
+		// Link products to existing catalog objects.
+		foreach ( $existing_catalog_objects as $remote_catalog_item ) {
+			if ( ! $remote_catalog_item instanceof CatalogObject ) {
+				continue;
+			}
+
+			if ( is_array( $remote_catalog_item->getItemData()->getVariations() ) ) {
+				$product_id = null;
+				foreach ( $remote_catalog_item->getItemData()->getVariations() as $catalog_item_variation ) {
+					$variation_data = $catalog_item_variation->getItemVariationData();
+					if ( ! $variation_data || ! $variation_data->getSku() || empty( $variation_data->getSku() ) ) {
+						continue;
+					}
+
+					$local_product_id = wc_get_product_id_by_sku( $variation_data->getSku() );
+					if ( ! $local_product_id ) {
+						continue;
+					}
+
+					$local_product = wc_get_product( $local_product_id );
+					if ( ! $local_product ) {
+						continue;
+					}
+
+					Product::update_square_meta(
+						$local_product,
+						array(
+							'item_variation_id'      => $catalog_item_variation->getId(),
+							'item_variation_version' => $catalog_item_variation->getVersion(),
+						)
+					);
+
+					$product_id = $local_product->is_type( ProductType::VARIATION ) ? $local_product->get_parent_id() : $local_product->get_id();
+				}
+
+				// Update the parent product if it exists.
+				if ( $product_id ) {
+					$product = wc_get_product( $product_id );
+					if ( $product ) {
+						Product::update_square_meta(
+							$product,
+							array(
+								'item_id'       => $remote_catalog_item->getId(),
+								'item_version'  => $remote_catalog_item->getVersion(),
+								'item_image_id' => Product::get_catalog_item_thumbnail_id( $remote_catalog_item ),
+							)
+						);
+					}
+				}
+
+				$linked_product_ids[] = $product_id;
+			}
+		}
+
+		return $linked_product_ids;
+	}
+
+	/**
+	 * Finds existing Square items by SKUs.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param array $product_skus The SKUs of the WooCommerce products to find.
+	 * @return array The IDs of the existing Square items.
+	 */
+	protected function find_existing_square_items_by_skus( $product_skus ) {
+		$existing_items = array();
+
+		$query     = new CatalogQuery();
+		$set_query = new CatalogQuerySet( 'sku', $product_skus );
+		$query->setSetQuery( $set_query );
+
+		try {
+			$response = wc_square()->get_api()->search_catalog_objects(
+				array(
+					'object_types'            => array( 'ITEM_VARIATION' ),
+					'query'                   => $query,
+					'include_related_objects' => true,
+				)
+			);
+
+			$search_response = $response->get_data();
+
+			if ( ! $search_response instanceof SearchCatalogObjectsResponse ) {
+				return array();
+			}
+
+			$related_objects = $search_response->getRelatedObjects();
+			if ( empty( $related_objects ) || ! is_array( $related_objects ) ) {
+				// No related objects found.
+				return array();
+			}
+
+			foreach ( $related_objects as $object ) {
+				if (
+					$object instanceof CatalogObject
+					&& $object->getType() === CatalogObjectType::ITEM
+					&& $object->getIsDeleted() === false
+				) {
+					$existing_items[] = $object;
+				}
+			}
+
+			return $existing_items;
+		} catch ( \Exception $exception ) {
+			wc_square()->log(
+				sprintf(
+					'Failed to find existing Square items by SKUs: %s',
+					$exception->getMessage()
+				)
+			);
+		}
+
+		return array();
+	}
 
 	/**
 	 * @throws \Exception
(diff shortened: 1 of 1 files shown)


includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+			return $linked_product_ids;
+		}
+
+		// Link products to existing catalog objects.
+		foreach ( $existing_catalog_objects as $remote_catalog_item ) {
+			if ( ! $remote_catalog_item instanceof CatalogObject ) {
+				continue;
+			}
+
+			if ( is_array( $remote_catalog_item->getItemData()->getVariations() ) ) {
```
faisal-alvi: I see chances of a Fatal null dereference on `getItemData()`

The `getItemData()` can return null if the catalog object is malformed. Although the result is wrapped in `is_array()`, PHP still evaluates `->getVariations()` first - before `is_array()` has a chance to run - so a null return from `getItemData()` can cause a fatal error.

**Fix:**
```suggestion
			$item_data = $remote_catalog_item->getItemData();
			if ( $item_data && is_array( $item_data->getVariations() ) ) {
```

And then using it below:
```
// ...
foreach ( $item_data->getVariations() as $catalog_item_variation ) {
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+			if ( empty( $related_objects ) || ! is_array( $related_objects ) ) {
+				// No related objects found.
+				return array();
+			}
+
+			foreach ( $related_objects as $object ) {
+				if (
+					$object instanceof CatalogObject
+					&& $object->getType() === CatalogObjectType::ITEM
+					&& $object->getIsDeleted() === false
```
faisal-alvi: The Square SDK's `getIsDeleted()` returns `null` when an object is active/not deleted. With strict === `false`, `null` !== `false` evaluates true, causing the condition to fail for all active objects. As a result, `$existing_items` is always empty, the guard becomes ineffective, and duplicates continue to be created.

So instead of a strict check, where null !== false, we can go for loose falsy check that handles `null` and `false`.

```suggestion
					&& ! $object->getIsDeleted()
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -745,6 +923,37 @@ protected function upsert_new_products() {
 			$this->set_attr( 'upsert_new_product_ids', array() );
 		}
 
+		// SKU guard: check for existing Square items before creating new ones.
+		// Link products to existing Square items by SKUs, to prevent creating duplicate items.
+		$linked_product_ids = $this->link_products_to_existing_square_items( $product_ids );
```
faisal-alvi, line 968: This is a suggestion to avoid unnecessary API calls. The SKU guard re-runs on timeout re-entry (wasted API calls, rate-limit risk).

If the job times out after `link_products_to_existing_square_items()` runs but before `set_attr( 'processed_product_ids' )` is persisted, on re-entry those linked IDs are not yet in `processed_product_ids`. The guard fires again - redundant API calls that can compound the original timeout.

We can add a fix by persisting linked IDs to a dedicated job attribute (for example, `upsert_linked_product_ids`) immediately after the guard runs, and subtract them on re-entry. Something like:

```
$already_linked_ids = $this->get_attr( 'upsert_linked_product_ids', array() );
$product_ids        = array_values( array_diff( $product_ids, $already_linked_ids ) );

$linked_product_ids = $this->link_products_to_existing_square_items( $product_ids );

if ( ! empty( $linked_product_ids ) ) {
    $already_linked_ids = array_merge( $already_linked_ids, $linked_product_ids );
    $this->set_attr( 'upsert_linked_product_ids', $already_linked_ids );
    // ... rest of the guard block
}
```
What are your thoughts?

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+		foreach ( $existing_catalog_objects as $remote_catalog_item ) {
+			if ( ! $remote_catalog_item instanceof CatalogObject ) {
+				continue;
+			}
+
+			if ( is_array( $remote_catalog_item->getItemData()->getVariations() ) ) {
+				$product_id = null;
+				foreach ( $remote_catalog_item->getItemData()->getVariations() as $catalog_item_variation ) {
+					$variation_data = $catalog_item_variation->getItemVariationData();
+					if ( ! $variation_data || ! $variation_data->getSku() || empty( $variation_data->getSku() ) ) {
```
faisal-alvi: This can be simplified:
```suggestion
					if ( ! $variation_data || empty( $variation_data->getSku() ) ) {
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+		// If no existing catalog objects found, return.
+		if ( empty( $existing_catalog_objects ) ) {
+			return $linked_product_ids;
+		}
+
+		// Link products to existing catalog objects.
+		foreach ( $existing_catalog_objects as $remote_catalog_item ) {
+			if ( ! $remote_catalog_item instanceof CatalogObject ) {
+				continue;
+			}
```
faisal-alvi: Every entry in `$existing_catalog_objects` was already guarded by `instanceof CatalogObject` when inserted into the map (see line 766 above). The re-check here in the linking loop can be removed.

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+		} catch ( \Exception $exception ) {
+			wc_square()->log(
+				sprintf(
+					'Failed to find existing Square items by SKUs: %s',
+					$exception->getMessage()
+				)
+			);
+		}
+
+		return array();
```
faisal-alvi, line 935: The catch block logs and returns [], causing the guard to be skipped and the upsert to proceed normally. I guess this is intentional (fail-open to not block the sync), but we should add an inline comment explaining the reasoning.

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+							)
+						);
+					}
+
+					$linked_product_ids[] = $product_id;
+				}
+			}
+		}
+
+		return $linked_product_ids;
```
faisal-alvi, line 875: Suggestion: should we add a log entry when products are linked, including the total number of linked product IDs?

includes/Sync/Manual_Synchronization.php
```diff
@@ -716,6 +719,181 @@ protected function search_matched_products() {
...
+	 */
+	protected function find_existing_square_items_by_skus( $product_skus ) {
+		$existing_items = array();
+
+		$query     = new CatalogQuery();
+		$set_query = new CatalogQuerySet( 'sku', $product_skus );
+		$query->setSetQuery( $set_query );
+
+		try {
+			$response = wc_square()->get_api()->search_catalog_objects(
```
faisal-alvi, line 894: I just realised that the `search_catalog_objects` call has the cursor-based pagination. Is it worth considering the pagination scenarios for this SKU check?

faisal-alvi (changes requested):
@iamdharmesh Thanks for working on the fix. I’ve left some feedback below for you to review.

**My main concern** is about the Duplicate Square SKUs:

If two Square items each have a variation with the same SKU, both parent ITEM objects will return by the API and be placed into `$existing_catalog_objects`. The linking loop will then call `wc_get_product_id_by_sku()` for both, finding the same WooCommerce product each time, and call `Product::update_square_meta()` twice. The second call will silently overwrite the first - the product will end up linked to whichever Square item is last in iteration order, with no warning.

The [spec](https://linear.app/a8c/issue/SQUARE-290/timeout-during-search-matched-products-causes-duplicate-square-catalog) requires:

> "If multiple Square items share the same SKU, the sync logs a warning and skips rather than guessing."

A potential fix would be: before the linking loop, build a `$sku -> [item_id, ...]` map. Any SKU that appears in more than one item can be considered a collision - we log a warning naming the conflicting Square item IDs and skip all variations with that SKU.

Another question (aka suggestion):

Should we also consider running the guard in `search_matched_products` too? The guard only covers `upsert_new_products`. A timeout mid-`search_matched_products` can also result in products with cleared meta falling through; the guard catches them on retry, but only after the search step completes in full.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Thanks for note @faisal-alvi.

I see this as a SKU guard only, and it's very unlikely that we will have lots of existing products that will match or cause the timeout. Even if that happens, it would be an edge case.

> If the job times out after link_products_to_existing_square_items() runs but before set_attr( 'processed_product_ids' ) is persisted

This applies to the suggested fix as well. Currently, we are setting the value to `processed_product_ids`, and the suggestion asks to set it to `upsert_linked_product_ids`. So, a timeout can occur before setting that too. Given that this is only a guard and there is only one edge case where we get lots of existing items from Square, which could result in a timeout, I would keep things as they are.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Removed.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Done.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Done.

includes/Sync/Manual_Synchronization.php
iamdharmesh: We are already sending request in 250 SKUs batch. So, no pagination needed here.

iamdharmesh:
Thanks for review @faisal-alvi.

All feedback are addressed now. Could you please check now?

> My main concern is about the Duplicate Square SKUs:
>
> If two Square items each have a variation with the same SKU, both parent ITEM objects will return by the API and be placed into $existing_catalog_objects. The linking loop will then call wc_get_product_id_by_sku() for both, finding the same WooCommerce product each time, and call Product::update_square_meta() twice. The second call will silently overwrite the first - the product will end up linked to whichever Square item is last in iteration order, with no warning.
>
> The [spec](https://linear.app/a8c/issue/SQUARE-290/timeout-during-search-matched-products-causes-duplicate-square-catalog) requires:
>
> "If multiple Square items share the same SKU, the sync logs a warning and skips rather than guessing."


Updated code to skip now and add alert for the same.

> Should we also consider running the guard in search_matched_products too? The guard only covers upsert_new_products. A timeout mid-search_matched_products can also result in products with cleared meta falling through; the guard catches them on retry, but only after the search step completes in full.

I don't see any need there. The upsert new products step starts just after the search step is completed. So, that is basically the same as what you have asked. Thanks.

includes/Sync/Manual_Synchronization.php
```diff
@@ -771,6 +772,37 @@ protected function link_products_to_existing_square_items( $product_ids ) {
 					continue;
 				}
 				$existing_catalog_objects[ $remote_catalog_object_id ] = $existing_catalog_object;
+
+				// Check for Multiple Square items with the same SKU.
+				$item_data = $existing_catalog_object->getItemData();
+				if ( $item_data && is_array( $item_data->getVariations() ) ) {
+					foreach ( $item_data->getVariations() as $variation ) {
+						if ( ! $variation->getItemVariationData()->getSku() ) {
```
faisal-alvi: I see a fatal error risk here. The `getItemVariationData()` can return `null` for a malformed API response. Chaining `->getSku()` on `null` can raise a fatal PHP error. The code below on line 820 already does this correctly. We should apply the same `null` guard here.

```
$variation_data = $variation->getItemVariationData();
if ( ! $variation_data || ! $variation_data->getSku() ) {
    continue;
}
$sku = $variation_data->getSku();
```

faisal-alvi (commented):
@iamdharmesh Thanks for addressing the feedback. Now I only see one point related to a fatal error. Please take a look and share your thoughts.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Thanks @faisal-alvi. Addressed in https://github.com/woocommerce/woocommerce-square/pull/486/commits/e9897d7e9e96bf06d59b4fa7f811ee29e77fdd41

faisal-alvi (approved):

ankitguptaindia:
### **QA Status:** ✅ Passed


###  **Test Results**

Tested the fix by simulating the case described in the issue.

The issue is reproducible on the `trunk` branch and results in duplicate product creation on the Square side. After applying the PR changes, the issue is resolved and no duplicate products are created.

* Duplicate product creation is reproducible on `trunk` (with simulated condition) 
* PR successfully prevents duplicate product creation 
* Reverting back to `trunk` confirms the issue is reproducible again 

---


### **Trunk branch (before fix): Issue persists**

image

---

### **PR branch (with fix): Issue resolved** 


image

---

### **Negative Test (reverted to trunk): Issue reproducible again** 

image

---

###  **Conclusion**

The fix works as expected. The PR successfully resolves the issue by preventing duplicate product creation when `_square_*` metadata is missing.

ankitguptaindia (approved):

ankitguptaindia:
------------

### **Testing Environment -** 


<Details>

- WordPress: 6.9.4
- PHP: 8.4.18
- Server: nginx/1.26.1
- Browser: Chrome 147.0.0.0
- OS: macOS
- Plugins:
  * WooCommerce 10.7.0


</Details>

### Testing Scope:
- Smoke test executed to validate major functionality.
- Regression checks performed on key user flows related to this task/PR. 

### Result:
Regression testing passed ✅

### Final Status:
Marked as Ready for Merge 🚀

PR #498, opened 2026-05-15
SQUARE-278: [UI Update] 1 - Establish New Settings Tab Shell & Remove Legacy Menu

faisal-alvi:
### All Submissions:

* [x] Does your code follow the [WooCommerce Sniffs](https://github.com/woocommerce/woocommerce-sniffs/) variant of WordPress coding standards?
* [ ] Have you written new tests for your changes, as applicable? Foundational Playwright smoke tests (modern hub loads / legacy loads flag-off / escape-hatch forces legacy) are deferred to https://linear.app/a8c/issue/SQUARE-286/update-ui-9-end-to-end-qa-settings-migration-and-integration.
* [ ] Have you successfully run tests with your changes locally?
* [ ] Will this change require new documentation or changes to existing documentation?

---

### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/SQUARE-278/ui-update-1-establish-new-settings-tab-shell-and-remove-legacy-menu.

This is **Ticket 1** of the Square settings redesign series. It establishes the hub shell and routing infrastructure that Tickets 2-5 build on. No settings fields are added here; only the tab structure, gating logic, and legacy redirects.

image

#### WooCommerce 11.0 requirement

The modern path requires **WooCommerce 11.0** (WC Core PR [#65975](https://github.com/woocommerce/woocommerce/pull/65975)), which adds `SettingsSection::get_settings_ui_page()`, the native registry hook Square uses to slot its hub into WooCommerce > Settings > Payments > Square.

The gate in `Plugin::is_modern_settings_path_active()` checks two things:
1. `method_exists(SettingsSection, 'get_settings_ui_page')` - WC 11.0 presence check
2. `Features::is_enabled('settings-ui')` - feature flag

On WC < 11.0 or when the flag is off, `Settings_Page.php` (legacy top-level Square tab) is used unchanged.

#### Note on `Features::is_enabled` vs `FeaturesUtil`

`FeaturesUtil::feature_is_enabled()` delegates to `FeaturesController`, which only covers features registered in `FeaturesController::get_feature_definitions()`. The `settings-ui` flag is a **WooCommerce admin JS feature** and lives in the `woocommerce_admin_features` filter system (same one that populates `window.wcAdminFeatures`), not in `FeaturesController`. Using `FeaturesUtil` for it always returns `false`. `Features::is_enabled()` is the correct check. A comment in `Plugin.php` documents this distinction.

### Steps to test the changes in this Pull Request:

[...]

PR #505, opened 2026-06-10
SQUARE-279: [UI Update] 2 - Migrate & Restyle: General Tab

faisal-alvi:
### All Submissions:

* [ ] Have you written new tests for your changes, as applicable? Foundational Playwright smoke tests are deferred to https://linear.app/a8c/issue/SQUARE-286/update-ui-9-end-to-end-qa-settings-migration-and-integration.

---

Closes https://linear.app/a8c/issue/SQUARE-279/ui-update-2-migrate-and-restyle-general-tab.

This is **Ticket 2** of the Square settings redesign series. It migrates the General tab to the WooCommerce settings UI SDK and restyles it to match the approved Figma design. Built on top of Ticket 1 (https://github.com/woocommerce/woocommerce-square/pull/498) which established the hub shell.

The General tab is rendered as two SDK card sections: "Connect to Square" (environment radio, sandbox credentials, OAuth Connect/Disconnect) and "Business location" (a native SDK `select` whose options are rendered server-side). Both card headings use the SDK group `title`/`description` rather than a custom header component. Sandbox credential fields appear only when the Sandbox radio is selected, via the SDK's `fieldVisibility` predicate. OAuth Connect is production-only and hides immediately when Sandbox is selected (it reads the live `values` prop); the Disconnect action shows whenever the store is connected. The Business location list is rendered server-side and repopulated on a post-save page reload — matching the legacy settings page (including its `ACTIVE` + `CREDIT_CARD_PROCESSING` location filter) and keeping the server as the source of truth.

Settings persist to the existing `wc_square_settings` option via the existing `/wc/v3/wc_square/settings` REST controller. No new option keys are introduced. The save flow uses the SDK's `custom` adapter; the Square save handler routes changed fields to the existing REST endpoint and throws a translatable Error on failure so the SDK renders a clean notice. The sandbox access token is (re)applied to the encrypted token store whenever it differs from the stored value, so switching into sandbox propagates the saved token (keeping the connection valid and the Business location section visible) without re-encrypting an unchanged token on every save.

#### WooCommerce 11.0 dependency for the modern path

This change requires **WooCommerce 11.0** (milestone for WC Core PR [#65975](https://github.com/woocommerce/woocommerce/pull/65975)), which adds native registry support (`SettingsSection::get_settings_ui_page()`) and finalised the SDK identifiers:

| Concept | v2 zip (pre-release) | WC 11.0 |
|---|---|---|
| Feature flag | `modern-settings` | `settings-ui` |
| `WC_Settings_Page` method | `get_modern_settings_page()` | `get_settings_ui_page()` |
| Interface | `ModernSettingsPageInterface` | `SettingsUIPageInterface` |
| Script handle / JS global | `wc-modern-settings-sdk` / `modernSettingsSdk` | `wc-settings-ui` / `settingsUi` |
| Layout | 2-column sidebar+panel | 1-column centered cards |

On WC < 11.0 or when the flag is off, the legacy `Settings_Page.php` path is used unchanged.

### Screenshots

image

**Environment:** WooCommerce 11.0 dev build (use the zip from dmallory, WC Core PR #65975 branch, until Beta 1 ships on July 13). Verify in browser console: `window.wcAdminFeatures['settings-ui']` is `true`.

**Setup - two mu-plugin files are needed. Drop them into `wp-content/mu-plugins/`:**
- [modern-settings-flag.php](https://gist.github.com/faisal-alvi/09a5996522d214c86d2d019155821d12) - enables `settings-ui` feature flag
- [square-disable-modern-settings.php](https://gist.github.com/faisal-alvi/6a3cef28b42fa1b02b81e18bf756f879) - escape hatch, forces Square back to legacy even when the flag is on

---

**Modern path - General tab render:**

1. Navigate to **WooCommerce > Settings > Payments > Square**. The hub loads with 4 sub-tabs, General active by default.
2. Confirm the General tab renders two cards: "Connect to Square" and "Business location". Compare layout against the Figma design.
3. The "Connect to Square" card shows the Environment Selection radio (Production / Sandbox) with sub-descriptions.
4. With **Sandbox** selected, the `SANDBOX APPLICATION ID` and `SANDBOX ACCESS TOKEN` input fields are visible.
5. With **Production** selected, the sandbox credential fields disappear (no page reload).
6. When connected, the "Business location" card shows a dropdown populated server-side from the connected Square account (active, card-processing locations only).

**Save flow:**

7. With Sandbox selected, change the Sandbox Application ID value, then click **Save** in the top right. A "Settings saved." notice appears and the page reloads. Confirm the change persisted.
8. Toggle the Environment radio to Production, click Save. Confirm `wc_square_settings.enable_sandbox` updates in the database. Toggle back to Sandbox, click Save again to restore.
9. Confirm a save error (simulate by temporarily blocking the REST endpoint) renders an inline error notice rather than failing silently.

**OAuth visibility and Business location:**

10. While disconnected in Production, the "Connect to Square" button is shown. Toggle to **Sandbox**: the Connect button hides immediately (OAuth is production-only) and the manual Sandbox credential fields appear. Toggle back to **Production**: the Connect button returns. No page reload.
11. After saving an environment change, the page reloads and the Business location dropdown is re-rendered server-side with that environment's available locations and the saved selection.

**Empty sub-tabs (out of scope for this ticket):**

12. Click each of Payment methods, Payments & Transactions, Synchronize Square sub-tabs. Each renders the hub shell with no crash and no Save button. These tabs are filled in by Tickets 3, 4, 5.

**Legacy preservation:**

13. Remove `modern-settings-flag.php`. The top-level **WooCommerce > Settings > Square** tab returns and the legacy General tab renders unchanged.
14. Re-add the flag mu-plugin. Add `square-disable-modern-settings.php` (escape hatch, uses filter `wc_square_is_modern_settings_active`). Confirm Square falls back to legacy even with the flag on. Remove the escape hatch to restore the modern hub.

### Changelog entry

Add - General settings tab migrated to the WooCommerce 11.0 settings UI SDK with a single-column card layout matching the approved Figma design.


faisal-alvi:
@iamdharmesh This is ready for your review.

src/settings/general/oauth-connect.js
```diff
@@ -0,0 +1,73 @@
+import { useState, useEffect } from '@wordpress/element';
+import { __ } from '@wordpress/i18n';
+import apiFetch from '@wordpress/api-fetch';
+
+export default function OAuthConnect( { values } ) {
```
iamdharmesh, line 23: I think this should be visible only when the production environment selected. Currently it's showing for both.

src/settings/general/location-picker.js
```diff
@@ -0,0 +1,115 @@
...
+		// Optimistic update.
+		setState( ( prev ) => ( {
+			...prev,
+			[ key ]: newId,
+			saving: true,
+			error: '',
+		} ) );
+
+		try {
+			await apiFetch( {
```
iamdharmesh: The location options are not getting updated as per the selection on the env. (even after saving it). Could you please check?

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -53,49 +50,181 @@ public function get_page_id(): string {
...
+		 * @return array<string, array>
+		 */
+		private function get_general_tab_groups(): array {
+			$settings = (array) get_option( 'wc_square_settings', array() );
+
+			return array(
+				'square_connect_section'  => array(
+					'id'          => 'square_connect_section',
+					'title'       => '',
+					'description' => '',
```
iamdharmesh: How about using this to avoid using custom component for the section header?
```suggestion
					'title'       => __( 'Connect to Square', 'woocommerce-square' ),
					'description' => __( 'Choose between connecting to a live production account for real transactions or a sandbox account for testing purposes.', 'woocommerce-square' ),
```

src/settings/general/location-picker.js
```diff
@@ -0,0 +1,115 @@
...
+				),
+			} ) );
+		}
+	};
+
+	return (
+		<div>
+			<select
+				value={ currentId }
+				onChange={ handleChange }
```
iamdharmesh: "Connect to Square" section fields are getting saved on click the save button, while this is on the selecting option. Could we make it common for all?

iamdharmesh (commented):
HI @faisal-alvi, Thanks for the PR. Great work here. I have added few initial comments could you please check when you get a chance? Thanks!

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -53,49 +50,172 @@ public function get_page_id(): string {
...
+				),
+			);
+
+			// Only show the Business location section once connected; there are no
+			// locations to choose before connecting.
+			if ( wc_square()->get_settings_handler()->is_connected() ) {
+				$groups['square_location_section'] = array(
+					'id'          => 'square_location_section',
+					'title'       => '',
+					'description' => '',
```
iamdharmesh: Same as Square connect section can we use this as well to get rid of customer section header component?

src/settings/general/section-header.js
```diff
@@ -0,0 +1,38 @@
...
+ * and forces a gray background that we don't want.
+ *
+ * Field schema:
+ *   - field.label       (string) — heading text
+ *   - field.description (string) — description HTML (rendered with RawHTML)
+ *
+ * @param {Object} props
+ * @param {Object} props.field
+ */
+export default function SectionHeader( { field } ) {
```
iamdharmesh: I think we can remove this.

src/settings/save-handler.js
```diff
@@ -0,0 +1,73 @@
...
+		throw new Error(
+			error?.message ||
+				__(
+					'Unable to save Square settings. Please try again.',
+					'woocommerce-square'
+				)
+		);
+	}
+
+	// Reload after saving so the Business location dropdown repopulates from the
```
iamdharmesh: As we are reloading the page to repopulate the locations, which we can provide options at the PHP side registration also. So, do we need the custom location picker now?

iamdharmesh: Same for environment selector as well? 🤔

includes/Admin/Rest/WC_REST_Square_Settings_Controller.php
iamdharmesh: As we are seeding the ` $settings` from the stored option now, this will be fires on every save in sandbox mode. Previously it was skipped unless the token was set.

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -53,49 +50,172 @@ public function get_page_id(): string {
...
+
+		/**
+		 * Returns the field groups for the General tab.
+		 *
+		 * @since x.x.x
+		 *
+		 * @return array<string, array>
+		 */
+		private function get_general_tab_groups(): array {
+			$settings = (array) get_option( 'wc_square_settings', array() );
```
iamdharmesh: Better to use `SQUARE_GATEWAY_SETTINGS_OPTION_NAME` const here. instead of hardcoded `wc_square_settings`

src/settings/general/environment-selector.js
```diff
@@ -0,0 +1,87 @@
...
+						htmlFor={ `square_environment_${ option.value }` }
+						aria-label={ option.label }
+						style={ {
+							display: 'flex',
+							alignItems: 'flex-start',
+							gap: '8px',
+							marginBottom: '12px',
+							cursor: 'pointer',
+							fontWeight: 'normal',
+						} }
```
iamdharmesh: As we already have modern-hub.css. How about moving this there?

src/settings/general/environment-selector.js
```diff
@@ -0,0 +1,87 @@
...
+ * @param {Function} props.onChange Called with the new value on change.
+ */
+export default function EnvironmentSelector( { field, value, onChange } ) {
+	return (
+		<div>
+			{ field?.label && (
+				<div style={ { fontWeight: 600, marginBottom: '12px' } }>
+					{ field.label }
+				</div>
+			) }
```
iamdharmesh, line 43: A11y: use <legend> for the group label of the following fieldset.

iamdharmesh (changes requested):
Thanks @faisal-alvi. This looks good now. However, we still need to address few things before mark this as a ready for the review. Could you please help to check on that?

includes/Admin/Rest/WC_REST_Square_Settings_Controller.php
```diff
@@ -192,9 +192,12 @@ public function get_settings() {
...
 		$filtered_settings['disconnection_url']      = html_entity_decode( wp_nonce_url( $url, $action ), ENT_QUOTES | ENT_SUBSTITUTE | ENT_HTML401 );
 
-		// Add locations to the response.
+		// Add locations to the response. Force a fresh fetch so the list reflects
+		// the currently connected environment instead of a stale transient (the
+		// hub runs under tab=checkout, so is_admin_settings_screen() is false and
+		// the cached transient would otherwise be served).
 		if ( wc_square()->get_settings_handler()->is_connected() ) {
-			$filtered_settings['locations'] = wc_square()->get_settings_handler()->get_locations();
+			$filtered_settings['locations'] = wc_square()->get_settings_handler()->get_locations( true );
```
iamdharmesh: Question:
As we are already requesting fresh locations in `get_location_options`. is this needed now?

src/settings/save-handler.js
iamdharmesh: > Same for environment selector as well? 🤔

I mean for custom components for environment selector as well.

src/settings/save-handler.js
```diff
@@ -0,0 +1,73 @@
...
+	// freshly fetched locations for the saved environment, matching the legacy
+	// settings page. The short delay lets the success notice render first. The
+	// SDK keeps a beforeunload guard active, so clear it first to avoid a "Leave
+	// site?" prompt blocking the programmatic reload.
+	setTimeout( () => {
+		window.onbeforeunload = null;
+		window.location.reload();
```
iamdharmesh: Question:
Do we need to reload for every save of env/crendentials change only?

src/settings/save-handler.js
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// `enable_sandbox` comes from a custom radio component (EnvironmentSelector)
+	// which calls onChange('yes') or onChange('no') directly — no conversion needed.
+	// The REST API already expects 'yes'/'no' and that is exactly what we have.
+
+	// The Business location select writes to a single `location_id` field; route
+	// it to the option key for the currently selected environment.
+	if ( 'location_id' in changedValues ) {
+		const isSandbox = values.enable_sandbox === 'yes';
```
iamdharmesh, line 30: Question:
Shouldn't this save the location value to wrong env key if environment value changed in save settings?

iamdharmesh (commented):
Thanks @faisal-alvi. This looks good now. However, I have added a few questions and I noticed that the save button gets stuck in the loading state if any settings options are not updated.(see attached video). Thanks


https://github.com/user-attachments/assets/fc4c21e7-b93e-4a50-84d0-615ea65d0114

src/settings/save-handler.js
faisal-alvi: Good catch. Location routing is now skipped when `enable_sandbox` is part of the same change set. After the env save reloads, the new environment's locations populate, and the merchant picks/saves the location on its own.

src/settings/save-handler.js
faisal-alvi: Agreed, scoped it. Reload only fires when an env/credential field (`enable_sandbox` / `sandbox_application_id` / `sandbox_token`) changed, since that's what repopulates the locations list. A location-only save no longer reloads.

includes/Admin/Rest/WC_REST_Square_Settings_Controller.php
faisal-alvi: Correct, removed it.

src/settings/save-handler.js
faisal-alvi: The location picker is already a native `select` fed by PHP options. For the environment selector I kept it custom on purpose: in the WC 11.0 `settings-ui` SDK a native radio field type doesn't render radio buttons - it maps to the same `SelectControl` dropdown as select (in `settings-ui` it's literally `"select" === type || "radio" === type → SelectControl`; there's no `RadioControl`). The Figma calls for actual Production / Sandbox radio buttons with a description line under each, which a dropdown can't express - so the custom component is warranted here, unlike the location picker.

faisal-alvi:
@iamdharmesh thanks for the feedback.

Regarding the Save button issue, it's a race in the WC 11.0 `settings-ui` SDK, not our code: after changing the environment and reverting it, the form is flagged dirty even though `values` is byte-identical to `initialValues`, and Save then hangs `isSaving` (the handler is never reached).

It reproduces even with `fieldVisibility` removed / a pure-CSS hide, and it's timing-sensitive (an async effect re-dirties the form ~1s after revert), so it isn't fixable extension-side. The Woo core team [confirmed](https://fueled.slack.com/archives/C05REAQUK28/p1782999947975739?thread_ts=1782999018.848949&cid=C05REAQUK28) it and is working on a fix, see https://github.com/woocommerce/woocommerce/issues/66204.

From our side, as discussed, I created a separate ticket for this, [SQUARE-344](https://linear.app/a8c/issue/SQUARE-344/save-button-stuck-in-loading-when-settings-unchanged-wc-settings-ui), to tackle later. We can skip this for now and pick it up in a separate PR once the SDK fix lands, rather than block the current PR.

Except for that one, I addressed other points from your feedback. Please have a look and let me know if any changes are required.

src/settings/save-handler.js
iamdharmesh: Thanks for the clarification.

iamdharmesh (approved):
Thanks @faisal-alvi. Looks good to me.

qasumitbagthariya:
QA update - Need feedback ⚠️
---
@faisal-alvi 

Everything looks good overall, but I noticed the [My Applications ↗](https://developer.squareup.com/apps) , [Sandbox Test Account ↗](https://developer.squareup.com/docs/devtools/sandbox/overview) , [Learn more about active locations ↗](https://squareup.com/help/us/en/article/5593-locations) links appear different compared to the legacy version. Could you please check this from your end?

image

qasumitbagthariya:
Thanks for the fix @faisal-alvi. Looks good to me.


https://github.com/user-attachments/assets/8f883c7d-e841-4097-ab07-38e84eebe9ea


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.0-dev
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.2 
- Branch: redesign/ticket-2-SQUARE-279

</Details>


🔁 Next Step  -  Ready for UAT

qasumitbagthariya (approved):

PR #506, opened 2026-06-14
Compatibility testing with WC 10.9

qasumitbagthariya:
### All Submissions:

* [ ] Does your code follow the [WooCommerce Sniffs](https://github.com/woocommerce/woocommerce-sniffs/) variant of WordPress coding standards?
* [ ] Have you written new tests for your changes, as applicable?
* [ ] Have you successfully run tests with your changes locally?
* [ ] Will this change require new documentation or changes to existing documentation?

---

### Changes proposed in this Pull Request:

Closes https://linear.app/a8c/issue/SQUARE-329/compatibility-testing-with-woo-109-beta-1

### Steps to test the changes in this Pull Request:

- Run the e2e test using GitHub action to test this PR
- All tests should be passed.
- Manually check plugin setup and plugin core functionality.

### Changelog entry

> Dev - Bump WooCommerce "tested up to" version 10.9.
> Dev - Bump WooCommerce minimum supported version to 10.7.

PR #507, opened 2026-06-19
Fix E2E stability timeouts by disabling CSS animations in Playwright config

faisal-alvi:
## Context

https://github.com/woocommerce/woocommerce-square/pull/506#issuecomment-4749181803

E2E `@sync` tests have been failing on CI. The suspected cause is Playwright 1.57+, which introduced stricter element stability checks: before interacting with an element, Playwright waits for its bounding box to stop changing (i.e., CSS transitions to complete). The WooCommerce settings page uses CSS transitions on form elements, which may cause `.click()` and `.check()` calls to wait indefinitely until timeout, meaning settings are never saved and sync never runs.

The `@playwright/test` dependency uses `^1.54.1` (unpinned minor/patch), so it can silently upgrade on CI.

## Approach

Added `reducedMotion: 'reduce'` to `tests/e2e/config/playwright.config.js`. This tells the browser to use the `prefers-reduced-motion: reduce` media query, which disables CSS animations and transitions globally across all tests. Playwright's stability checks then pass immediately since elements no longer animate.

This is a single-line config change with no modifications to individual test files.

## Status

This is a test PR to validate whether the `reducedMotion: 'reduce'` approach resolves the failures. Merging is pending CI confirmation.

Closes https://linear.app/a8c/issue/SQUARE-338/fix-e2e-sync-failures-caused-by-playwright-157-element-stability

### DIFF
--- a/tests/e2e/config/playwright.config.js
+++ b/tests/e2e/config/playwright.config.js
@@ -64,6 +64,8 @@ module.exports = defineConfig({
     baseURL: 'http://localhost:8889',
     storageState: 'state.json',
     navigationTimeout: 60000,
+    // Disable CSS animations/transitions so Playwright doesn't time out on @wordpress/components buttons.
+    reducedMotion: 'reduce',
   },
 
   /* Configure projects for major browsers */
@@ -105,4 +107,3 @@ module.exports = defineConfig({
   //   port: 3000,
   // },
 });
-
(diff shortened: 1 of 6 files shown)


iamdharmesh (approved):
Thanks for fixing this @faisal-alvi. Looks good to me. No QA needed as this is test config change only.

Issue #149, opened 2024-06-07
[Post-NUX] Disabling Credit Cards should not disable Digital Wallets


faisal-alvi:
### Describe the bug

Disabling Credit Cards should not disable Digital Wallets.

### To reproduce

n/a

**Screenshots**

### Expected behavior

**Environment (please complete the following information):**
 - WordPress Version
 - WooCommerce Version
 - WooCommerce Square Plugin Version
 - Browser [e.g. chrome, safari] and Version
 - Any other plugins installed

### Additional details

<details><summary>System status</summary>
  
  ```
  <!--If applicable, paste the system status here. Please ensure you redact or remove any identifying information. -->
  ```
</details>

Issue #102, opened 2024-04-15
[Post-NUX] Digital Wallet as a standalone Gateway


faisal-alvi:
### Describe the bug

The goal is to separate the Digital Wallet settings from the “Square Credit Card” payment gateway and make it as an individual payment gateway itself. Refer NUX figma from [PT](https://10upwoo.wordpress.com/2024/01/30/pt-nux-improvements-for-square/).

**Screenshots**

![image](https://github.com/woocommerce/woocommerce-square/assets/25176325/95e2a34b-99ae-4d9a-a263-9f24947f54a6)

### Additional details

Related #89

PR #514, opened 2026-07-01
SQUARE-280: [Update UI] 3 - Migrate & Restyle: Payment Methods Tab

faisal-alvi:
### All Submissions:

---

Square's payment methods (Credit Card, Digital Wallets, Cash App, and Gift Cards) each lived on their own separate settings screens, so a store manager had to move between several places to manage what is really one set of payment options.

This PR consolidates all four payment methods into a single Payment Methods tab in the modern Square settings hub (`WooCommerce > Settings > Payments > Square > Payment Methods`) on WooCommerce 11.0+. Each method has an enable/disable toggle, and all changes on the tab (including the toggles) are persisted together when you click **Save**. The display settings for Digital Wallets and Cash App now live in a "Customize" sub-view for that method, with live button previews rendered by the Square Web Payments SDK. The tab follows the approved Figma design.

It also adds a safety redirect: when the modern settings path is inactive (older WooCommerce, or the feature flag is off), any modern hub URL is redirected to the legacy Square settings tab so those links never land on an empty section.

The Digital Wallets behavioral change (splitting the "Hide Digital Wallet Buttons" multi-select into separate Google Pay / Apple Pay controls) is implemented: each wallet has its own enable toggle, button label, and button colour, persisted to per-wallet option keys on the Credit Card gateway. Coupling rules match the approved design. Enabling the parent defaults both wallets on; disabling both wallets turns the parent off; and each wallet's preview/Customize follows its own enable state. Bringing the **legacy** settings screen and the checkout/gateway rendering into line with these per-wallet keys is tracked separately in [SQUARE-346](https://linear.app/a8c/issue/SQUARE-346).

Closes https://linear.app/a8c/issue/SQUARE-280/update-ui-3-migrate-and-restyle-payment-methods-tab

### Notes for QA/reviewers

The header **breadcrumb** (used to navigate back from a Customize sub-view to the payment-methods list) is out of scope for this ticket and is handled separately ([Issue #6](https://linear.app/a8c/issue/SQUARE-283/update-ui-6-implement-new-settings-header-and-breadcrumb-components)). Until then, to return from a Customize sub-view use the browser **Back** button, or click the **Payments** tab. The Customize sub-views are URL-addressable (`&pm-view=digital-wallet` / `&pm-view=cash-app`), so they survive a reload and work with Back/Forward.

**Button label / colour options mirror the legacy settings screen, with one Google Pay caveat.** Apple Pay's button-label select offers the legacy options only: Buy Now (`buy`), Donate (`donate`), No Text (`plain`). Button colours match legacy too (Apple Pay: Black / White / White with outline; Google Pay: Black / White).

The Google Pay button-label select is limited to the two types the Square Web Payments SDK actually renders: `long` (full "Buy with G Pay") and `short` (mark only). The Google-API named labels (Buy now / Pay / Plain / …) are **not** accepted by `googlePay.attach()` and do not render in the preview or at checkout, so they are intentionally excluded. Legacy has no Google Pay label control today (the checkout button is hardcoded to "Buy with GPay"); adding one there, and wiring the checkout/gateway rendering to honour these per-wallet keys, is tracked in [SQUARE-346](https://linear.app/a8c/issue/SQUARE-346).

**Known issue: Google's dynamic button rendering.** When the viewer is signed in to Google with an eligible saved card, the Google Pay button re-renders from the plain "Buy with G Pay" label into Google's personalised button showing the card network + last four digits, shortly after the SDK loads. This looks like the selected label briefly "reverting". It is **Google's own button behaviour**, driven from Google's end (not the plugin), and it happens identically at checkout; the admin preview is faithful to what a signed-in shopper sees. This is documented by Google: [Google Pay Web brand guidelines](https://developers.google.com/pay/api/web/guides/brand-guidelines) ("When users have an eligible card on their Google Pay account, the 'Buy', 'Pay' and 'Plain' payment buttons will display the card network and the last four digits of the card number."). We keep the label control as-is and flag this as expected, known behaviour rather than a bug to fix.

**Apple Pay preview: could not be tested.** Apple Pay is not available in India, so the Apple Pay button preview and flow could not be verified live. The implementation is complete and has been code-reviewed and looks correct, but the Apple Pay button preview/behaviour needs a reviewer in a supported region to confirm.

### Payment Methods Tab

payment-methods-tab

### Digital Wallets Tab

digital-wallet-tab

### Cash App Tab

cash-app-tab

1. On WooCommerce 11.0+, go to `WooCommerce > Settings > Payments > Square > Payment Methods` and confirm all four methods (Credit/debit card, Digital wallet, Cash App, Square gift cards) render with an enable toggle.
2. Toggle one method off, click Save, reload the page, and confirm the state persisted.
3. Click "Customize" on Digital wallet, toggle Google Pay / Apple Pay, change their button labels and colours, click Save, reload, and confirm the enable states and values persisted. Confirm disabling both wallets turns the parent Digital wallet toggle off.
4. Click "Customize" on Cash App, change the Button Theme and Button Shape, click Save, reload, and confirm those values persisted.
5. Confirm the live Google Pay / Apple Pay / Cash App previews reflect the selected options.
6. Disable the modern settings path (older WooCommerce or the `settings-ui` feature off), open a modern hub URL such as `?page=wc-settings&tab=checkout&section=square&square-tab=payment-methods`, and confirm you are redirected to the legacy Square tab (`?page=wc-settings&tab=square`).

### Changelog entry

> Add - Consolidate Square payment methods (Credit Card, Digital Wallets, Cash App, Gift Cards) into a single Payment Methods tab in the modern settings hub, each with a Customize sub-view and live button previews.


src/settings/index.js
```diff
@@ -1,20 +1,59 @@
...
+import CashAppButtonPreview from './cash-app/button-preview';
+import DigitalWalletPreview from './digital-wallets/preview';
 import squareSaveHandler from './save-handler';
 
 registerSettingsExtension( {
 	scope: { page: 'square' },
 	components: {
 		'square/oauth-connect': OAuthConnect,
 		'square/environment-selector': EnvironmentSelector,
+		'square/section-header': SectionHeader,
```
iamdharmesh, line 17: We should remove this, if we are no longer using this.

src/settings/general/section-header.js
```diff
@@ -0,0 +1,40 @@
...
+ * not a title, and forces a gray background).
+ *
+ * Field schema:
+ *   - field.label       (string) — heading text
+ *   - field.description  (string) — optional description HTML (rendered with RawHTML)
+ *
+ * @param {Object} props
+ * @param {Object} props.field
+ */
+export default function SectionHeader( { field } ) {
```
iamdharmesh, line 20: Should be removed, as we are no longer using this.

src/settings/save-handler.js
```diff
@@ -28,22 +85,37 @@ export default async function squareSaveHandler( { values, changedValues } ) {
...
+				apiFetch( { path, method: 'POST', data: payloads[ path ] } )
+			),
+			...gatewayPuts.map( ( g ) =>
+				apiFetch( {
+					path: `/wc/v3/payment_gateways/${ g.id }`,
+					method: 'PUT',
+					data: { enabled: g.enabled },
+				} )
+			),
+		] );
```
iamdharmesh: `Promise.all` fires all requests concurrently. So, update get lost when same option updated twice eg:  `woocommerce_square_credit_card_settings`.

To reproduce: on the Payment Methods list, toggle Credit/debit card off and Digital wallet off, then Save, `enable_digital_wallets` and `enabled` are racing for the same array.

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -255,6 +262,306 @@ private function get_location_options(): array {
...
+			// Each payment method's enable state is a shared, reactive value read
+			// and written by the gateway list toggles (and, for digital wallets,
+			// the per-wallet toggles on the Customize sub-page). Nothing here
+			// self-saves: the values ride the SDK form and are persisted only when
+			// the page Save button is clicked (routed by save-handler.js), so the
+			// save adapter is 'none' on each.
+			$enable_state_fields = array(
+				'square_credit_card_enabled'  => $credit_card['enabled'] ?? 'yes',
+				'enable_digital_wallets'      => $credit_card['enable_digital_wallets'] ?? 'yes',
+				'square_cash_app_pay_enabled' => $cash_app['enabled'] ?? 'yes',
```
iamdharmesh: We are using  defaults `square_credit_card_enabled`, `square_cash_app_pay_enabled`, and `gift_cards_pay_enabled` to 'yes' when the option key is absent. But the gateway default is 'no' — see `Payment_Gateway.php:1195 ('default' => 'no')` and `Gift_Card.php:35 (get_option('enabled', 'no')`). On a store where a gateway has never been saved, the tab shows the toggle on while the gateway is actually off. Worse, because the save handler only sends `changedValues`, the merchant can't fix it by clicking Save as nothing is dirty, so nothing is sent. 

These three should default to 'no', or better, read `wc_square()->get_gateway( $id )->is_enabled()` directly.

src/settings/payment-methods/digital-wallet-toggle.js
```diff
@@ -0,0 +1,51 @@
...
+			checked={ value === 'yes' }
+			__nextHasNoMarginBottom
+			onChange={ ( checked ) => {
+				onChange( checked ? 'yes' : 'no' );
+
+				// Both wallets now off → disable the parent method (otherwise it
+				// would be "on" with nothing to display). Stay on this page.
+				const siblingOn = values?.[ siblingId ] === 'yes';
+				if ( ! checked && ! siblingOn ) {
+					setValue( 'enable_digital_wallets', 'no' );
```
iamdharmesh, line 50: We should enabled this back as well. If any one wallet get enabled.

includes/Admin/Payments_Square_Hub.php
```diff
@@ -112,6 +112,34 @@ public static function redirect_legacy_square_settings_tab(): void {
...
+	 * Redirects modern Square hub URLs to the legacy Square settings tab.
+	 *
+	 * Only hooked when the modern-settings path is inactive (WC < 11.0 or the
+	 * feature flag is off). In that state the hub section is not registered, so
+	 * ?tab=checkout&section=square renders an empty screen. Any such request
+	 * (including specific square-tab sub-tabs) is sent to the legacy Square tab.
+	 *
+	 * @since x.x.x
+	 */
+	public static function redirect_modern_hub_to_legacy(): void {
```
iamdharmesh, line 125: We should early return in case of AJAX endpoint called. Something like this, what do you think?
```php
if ( wp_doing_ajax() || ! current_user_can( 'manage_woocommerce' ) ) {
    return;
}
```

includes/Admin/Rest/WC_REST_Square_Credit_Card_Payment_Settings_Controller.php
```diff
@@ -150,6 +154,26 @@ public function register_routes() {
 						'type'              => 'array',
 						'sanitize_callback' => '',
 					),
+					'digital_wallets_google_pay_enabled'  => array(
+						'description'       => __( 'Whether the Google Pay button is enabled.', 'woocommerce-square' ),
+						'type'              => 'string',
+						'sanitize_callback' => '',
```
iamdharmesh: `'sanitize_callback' => ''` will  explicitly disables WordPress's default sanitization.  There's also no `validate_callback` and no enum, so This will happily persist any string.
I think we should add 'enum' to all four (yes|no, buy|donate|plain, and the colour sets) plus `'sanitize_callback' => 'sanitize_key'`.

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -255,6 +262,306 @@ private function get_location_options(): array {
...
+					'title'       => __( 'Choose your payment methods', 'woocommerce-square' ),
+					'description' => '',
+					'actions'     => array(),
+					'order'       => 0,
+					'fields'      => $enable_fields,
+				),
+				'digital_wallets_section' => array(
+					'id'          => 'digital_wallets_section',
+					'title'       => __( 'Digital wallet settings', 'woocommerce-square' ),
+					'description' => __( 'Allow customers to pay with Apple Pay or Google Pay from your Product, Cart and Checkout pages. <a href="https://woocommerce.com/document/woocommerce-square/payment-settings/enabling-digital-wallets/" target="_blank">Learn more about digital wallets</a>', 'woocommerce-square' ),
```
iamdharmesh: Better to use `sprintf()` with a %s placeholder for link. so translators can't break the markup

src/settings/index.js
```diff
@@ -1,20 +1,59 @@
...
+		'square/gateway-list': GatewayList,
+		'square/gateway-toggle': GatewayToggle,
```
iamdharmesh: Are we using this?

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -255,6 +262,306 @@ private function get_location_options(): array {
...
+		private function get_digital_wallet_preview_data(): string {
+			$is_sandbox = wc_square()->get_settings_handler()->is_sandbox();
+
+			return (string) wp_json_encode(
+				array(
+					'applicationId' => wc_square()->get_gateway()->get_application_id(),
+					'locationId'    => wc_square()->get_settings_handler()->get_location_id(),
+					'squareJsUrl'   => $is_sandbox
+						? 'https://sandbox.web.squarecdn.com/v1/square.js'
+						: 'https://web.squarecdn.com/v1/square.js',
```
iamdharmesh: Can we move this get JS CDN URL to shared helper function? I think we are using this hardcoded multiple places.

src/settings/cash-app/button-preview.js
```diff
@@ -0,0 +1,81 @@
...
+
+	const theme = values?.button_theme ?? 'dark';
+	const shape = values?.button_shape ?? 'semiround';
+	const themeClass = theme === 'dark' ? 'is-dark' : 'is-light';
+	const shapeClass = SHAPE_CLASS[ shape ] ?? SHAPE_CLASS.semiround;
+
+	return (
+		<div className="wc-square-preview">
+			<button
+				disabled
```
iamdharmesh, line 70: Lets add `type="button"` to prevent default type.

iamdharmesh (changes requested):
Thanks @faisal-alvi. This looks good overall. I haven't fully tested this but noticed few issue by reading code. Could you please help checking on those? Thanks.

src/settings/index.js
faisal-alvi: This one's still in use. The `square/section-header` renders the "Google Pay" and "Apple Pay" sub-headings in the Digital Wallet Customize view (`Square_Modern_Settings_Page.php`). Left it in.

src/settings/general/section-header.js
faisal-alvi: Kept. The `SectionHeader` component backs the Google Pay / Apple Pay sub-headings in the DW Customize view, so it's still wired up. Happy to revisit if you meant something else.

src/settings/save-handler.js
faisal-alvi: Fixed. The controller POSTs and the gateway-enable PUTs now run in two sequential steps instead of one `Promise.all`. A gateway PUT and a controller POST can target the same stored option (e.g. the `credit-card` gateway enabled vs `enable_digital_wallets`, both in `woocommerce_square_credit_card_settings`), so firing them together clobbered one write. Sequencing removes the race; each group still runs in parallel internally since those targets are distinct.

src/settings/payment-methods/digital-wallet-toggle.js
faisal-alvi: Fixed. Turning on either wallet now sets `enable_digital_wallets` = 'yes', symmetric to the both-off → parent-off rule, so enabling a wallet can't leave the parent stranded off.

src/settings/index.js
faisal-alvi: Good catch. The `square/gateway-toggle` was unused after a few adjustments. Removed the import, the registration, and deleted `gateway-toggle.js`.

src/settings/general/section-header.js
iamdharmesh: Got it. Thanks.

includes/Admin/Rest/WC_REST_Square_Credit_Card_Payment_Settings_Controller.php
iamdharmesh: Thanks for fixing this @faisal-alvi. I just noticed we have a similar issue in existing settings as well. However, that is not part of this PR. So, not a blocker for this, but we should fix those in the future. Thanks.

iamdharmesh (approved):
Thanks @faisal-alvi. This looks good now. Regarding Apple Pay preview, could you please add a note to get it QAed in the regression testing? So, we don't miss that. Thank you.

qasumitbagthariya (commented):
QA Update ❌
--

@faisal-alvi 

When the Credit/Debit Card setting is turned OFF, the Digital Wallet (Google Pay) and Square Gift Card options are also automatically turned OFF on the FE, even if they are enabled in the settings.


https://github.com/user-attachments/assets/ecb8d31e-0209-435e-bf67-8e48582e7eaa


Note:
I have not tested Apple Pay because it is not available in my region.


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.1
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.0-dev
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: redesign/ticket-3-SQUARE-280

</Details>

faisal-alvi:
@qasumitbagthariya Thanks for the report.

This is expected, pre-existing behavior, and it's actually already been raised and decided on:

- #149 → closed as not planned
- #102 → closed as not planned

**Digital Wallets** (Google Pay / Apple Pay) have no independent payment rail, they process through the Square Credit/Debit Card gateway, so when Credit/Debit Card is disabled they genuinely can't function and are hidden on the storefront:

- `Digital_Wallet::is_available()` requires the card gateway to be available (`… && $this->gateway->is_available()`).

**Square Gift Cards** are actually a standalone gateway (`Gift_Card extends Payment_Gateway`, enabled independently). The only coupling is that the checkout gift-card redemption field's script bails when Credit/Debit Card is off:

- `Gift_Card::enqueue_scripts()` returns early unless the card gateway is enabled, so the redemption UI doesn't load on the storefront.

Both checks pre-date this redesign and behave identically on the current legacy settings screen (this PR never touches `Gift_Card.php` or the checkout logic). The redesign project only focuses on the design aspects, not functionality tweaks, so this PR just adds the new toggle UI.

Making Digital Wallets behave fully independently of Credit/Debit Card is a significant architectural change (they'd need their own payment rail) and would warrant its own small project rather than being folded into this PR or the redesign work.

The one fair point is UX: the new Payment Methods tab makes the four methods look independent, when Digital Wallet (hard dependency) and the Gift Card redemption field (script coupling) both currently rely on Credit/Debit Card being on.

@builtbytay @iamdharmesh since true decoupling is out of scope here (#149/#102), the cleanest UX fix would be to **disable the Digital Wallet area behind a greyed-out overlay when Credit/Debit Card is turned off**, with a short note explaining the dependency, rather than letting it look independently toggleable. The Gift Card coupling is narrower (just the redemption-field script) and may be worth a separate look. Does that work, or would you prefer we leave it as-is? Happy to open a small follow-up ticket if needed.

iamdharmesh:
> @builtbytay @iamdharmesh since true decoupling is out of scope here (#149/#102), the cleanest UX fix would be to **disable the Digital Wallet area behind a greyed-out overlay when Credit/Debit Card is turned off**, with a short note explaining the dependency, rather than letting it look independently toggleable. The Gift Card coupling is narrower (just the redemption-field script) and may be worth a separate look. Does that work, or would you prefer we leave it as-is? 

Thanks for explaining @faisal-alvi. I think yes disable the Digital Wallet area behind a greyed-out overlay is the way to go here. Let's proceed with that cc: @builtbytay.

faisal-alvi:
@iamdharmesh I've disabled the Digital Wallet area behind a greyed-out overlay. Please have a look.

image

iamdharmesh (approved):
Thanks for the changes, @faisal-alvi. Disabling/greying out digital wallets when credit card payments are disabled looks good to me.

qasumitbagthariya:
QA Update: Need Feedback ⚠️
--

@faisal-alvi 

I noticed that the black border configured for the Cash App in the backend settings is not reflected on the frontend. The issue is reproducible on both themes(storefront,TT5).

image

Classic checkout
image

Block checkout
image

faisal-alvi:
Thanks @qasumitbagthariya. The Light theme was actually saving and applying correctly - Square's Cash App Pay SDK renders that theme as a plain white button with no outline, and there is no bordered variant in the SDK (both the new and legacy settings only offer Dark and Light).

The bug was in our settings preview, which drew a 1px black border around the light button and so promised an outline the real button never has. Fixed in https://github.com/woocommerce/woocommerce-square/pull/514/commits/9e55cb843ef2a8671492c82016ab159b0f3b2c94 - the preview now matches what the SDK renders. Please re-verify: set the theme to Light and confirm the preview shows a plain white button matching checkout, and that Dark still renders the solid black button on both classic and block checkout.

qasumitbagthariya:
QA Update: ✅
---

Thanks for the detailed comment @faisal-alvi 

Other than that, everything looks good to me. 

I have verified this PR in the `redesign/ticket-3-SQUARE-280` branch, which has been fixed and is functioning as intended.

- Payment Methods Tab
- Digital Wallets Tab
- Cash app tab

🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: redesign/ticket-3-SQUARE-280

</Details>


🔁 Next Step  -  Ready for UAT

qasumitbagthariya (approved):

PR #526, opened 2026-07-30
Automatically recover a stuck background sync job and unblock the queue

faisal-alvi:
### All Submissions:

---

When a background sync job dies mid run (timeout, fatal error, or a killed worker) its status stays "processing" forever. That stuck job blocks every later sync, and the healthcheck cannot restart the queue because pending or failing runner actions keep its guard condition satisfied. Nothing in the admin tells the merchant anything is wrong. One merchant ran this way for about a month and accumulated over 14,000 failed scheduler actions, discovered only by inspecting the database.

This change makes the sync self healing. Every processed step now records a heartbeat on the job, so a genuinely long running sync on a large catalog is never mistaken for a stuck one. The healthcheck looks for a job sitting in "processing" with no step activity beyond a threshold (15 minutes by default, filterable via `wc_square_stuck_job_threshold`). When it finds one it marks that job failed and releases the process lock so the next sync can run. Scheduled runner actions are deliberately left in place, so any other job waiting in the queue keeps processing. When more than one job is stalled, the oldest is recovered first, since that is the one holding up the queue.

Before failing anything it looks for signs of a live worker, because a false positive here would kill a working sync. Three signals are checked: the process lock, a runner action last touched within the threshold, and a runner action still pending. The last case earns a grace window rather than an immediate recovery (a third of the threshold by default, filterable via `wc_square_stuck_job_grace_period`), because a quiet site can cross the threshold with the queue merely paused and then resume on the very visit that triggered this check. At the defaults a paused queue is left alone for 15 minutes and a genuinely dead one is failed after 20. The recency bound on the second signal matters: an in progress row is only cleared by Action Scheduler's own cleaner, which runs from its queue runner, so on a site where Action Scheduler is not executing (the incident this exists for) an unbounded check would block recovery forever and rebuild the same deadlock in a new shape.

Because the reported incident also had the scheduled healthcheck itself failing, admin page loads act as a fallback trigger, throttled to once every five minutes and limited to users who can manage WooCommerce, so a site with broken cron heals as soon as someone visits the dashboard. That fallback deliberately does not run the whole healthcheck: it runs the recovery and the failed action cleanup, and restarts the queue only when it actually failed a job. Running the full healthcheck there would make every admin page load a third enqueue trigger, and since `handle()` takes the process lock without checking it first, that would widen the window for two runners to work the same step.

After a recovery a dismissible admin notice tells the merchant that a stuck sync was detected and stopped, that product data may be out of date, and links to the update page to start a new one. It clears as soon as a sync the merchant started begins processing, so a long sync does not keep showing a warning about the previous one, and it also clears if a sync has completed since the recovery. Interval polls are excluded from that dismissal because they start on their own every few minutes and would clear the notice before anyone read it. Recovery goes through the normal failure path, so the existing failed sync record and the (opt in, disabled by default) sync notification email behave exactly as they do for any other failed sync. Old failed scheduler actions from the plugin's sync hooks are also pruned daily after a retention period (30 days by default, filterable via `wc_square_failed_action_retention_days`).

The status panel and the stale sync warning shipped separately in SQUARE-299; this change is only the recovery mechanism the parent ticket asked for.

Closes https://linear.app/a8c/issue/SQUARE-291/a-stuckfailed-sync-job-blocks-all-subsequent-syncs

Setup: a site with WooCommerce Square connected (sandbox is fine), product sync enabled, and WP CLI available. Steps 1 to 4 reproduce the bug on trunk; steps 5 to 10 confirm the fix on this branch; steps 11 and 12 are the regressions that matter most.

Optional but recommended, so nothing needs a 15 minute wait: add this snippet to a mu plugin or to the theme's functions.php for the duration of testing, which shortens the stall threshold to one minute and the grace window to the enforced minimum of one minute.

```php
add_filter( 'wc_square_stuck_job_threshold', function () { return 60; } );
```

1. On trunk, simulate a stuck sync by running this WP CLI command. It injects a sync job that looks like it died two hours ago, then confirms the injection took.
   `wp eval '$h=wc_square()->get_background_job_handler();$h->clear_all_jobs();update_option("wc_square_background_sync_job_STUCKQA",wp_json_encode(array("id"=>"STUCKQA","status"=>"processing","action"=>"sync","manual"=>true,"started_processing_at"=>gmdate("Y-m-d H:i:s",time()-7200),"last_activity_at"=>time()-7200,"product_ids"=>array())));echo $h->get_job("STUCKQA") ? "injected" : "injection failed, run again";'`
   If it prints anything other than "injected", run it again. A sync or interval poll running in the background can consume the injected job.
2. In wp admin go to WooCommerce then Settings then Square then Update. The page shows "Syncing now..." even though nothing is running. This also confirms the injected job is really there.
3. Wait a few minutes (or run `wp eval 'wc_square()->get_background_job_handler()->handle_sync_healthcheck();'`), then reload the page. On trunk the job is still stuck, no notice appears anywhere, and Sync now stays unavailable. This is the bug.
4. Clean up: WooCommerce then Status then Tools then "Clear Square Sync".
5. Switch to this branch and repeat step 1 to inject the same stuck job.
6. Load any wp admin page (or run the healthcheck command from step 3). Recovery runs automatically.
7. Reload the Square Update page. The stuck "Syncing now..." state is gone and the Sync now button is available again. The Sync records list shows a new Failed entry reading "A sync stopped responding and was stopped automatically. Product data may be out of date, please start a new sync."
8. A dismissible admin notice appears reading "A stuck Square sync job was detected and automatically stopped. Product data may be out of date, please start a new sync from the update page." It shows on any admin screen, not only the Square pages.
9. Start a manual sync from the update page. The notice disappears as soon as that sync starts, without waiting for it to finish.
10. With logging enabled (WooCommerce then Status then Logs, square handle) the recovery writes "Auto-failed a stalled sync job (STUCKQA). The queue lock was released so the next sync can run."
11. Regression, a real sync must never be recovered out from under itself: start a normal manual sync on a catalog large enough to run past the stall threshold and let it finish. It must complete normally and never be flagged as stuck, because every step refreshes the job heartbeat. Worth running with the shortened threshold from the snippet above, which is a much harsher test than the 15 minute default.
12. Regression, a paused queue on a quiet site must not be treated as a dead one: inject the stuck job again as in step 1, then queue a runner action that is not due yet with `wp eval 'as_schedule_single_action(time()+3600,"wc_square_job_runner");echo "queued";'`. Run the healthcheck. The job must still be "processing" (the Update page still shows "Syncing now...") and the log records "Stalled sync has a queued runner action; allowing N seconds for the queue to resume before auto-failing." (N is 300 at the defaults, or 60 with the snippet above). Only once that window has passed does a later healthcheck fail the job.

### Changelog entry

> Fix - Automatically detect and recover a background sync job stuck in processing, unblock the sync queue, show an admin notice prompting a re run, and prune old failed scheduler actions.

### DIFF
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -109,6 +109,9 @@ class Admin {
 
 		// Show warning notice for the user to manually trigger a sync if it has been a while since the last successful sync.
 		add_action( 'admin_notices', array( $this, 'maybe_show_sync_status_notice' ) );
+
+		// Show a notice when a stalled sync job was automatically recovered, prompting a re-run.
+		add_action( 'admin_notices', array( $this, 'maybe_show_sync_recovered_notice' ) );
 	}
 
 
@@ -454,6 +457,61 @@ class Admin {
 						'<a href="' . esc_url( admin_url( 'admin.php?page=wc-settings&tab=square&section=update' ) ) . '">',
 						'</a>',
 						'<a href="' . esc_url( admin_url( 'admin.php?page=wc-status&tab=tools' ) ) . '">',
+						'</a>'
+					),
+					array(
+						'a' => array(
+							'href' => array(),
+						),
+					)
+				);
+				?>
+			</p>
+		</div>
+		<?php
+	}
+
+	/**
+	 * Maybe show a notice that a stalled sync job was automatically recovered.
+	 *
+	 * Shown when auto-recovery has fired and no successful sync has completed since, so the
+	 * merchant knows the previous sync was interrupted and restarted and can verify their products.
+	 * Clears itself once a sync completes successfully after the recovery.
+	 *
+	 * @since x.x.x
+	 *
+	 * @return void
+	 */
+	public function maybe_show_sync_recovered_notice() {
+		if ( ! current_user_can( 'manage_woocommerce' ) ) { // phpcs:ignore WordPress.WP.Capabilities.Unknown
+			return;
+		}
+
+		$settings = wc_square()->get_settings_handler();
+		if ( ! $settings->is_connected() || ! $settings->get_location_id() || ! $settings->is_product_sync_enabled() ) {
+			return;
+		}
+
+		$recovered_at = (int) get_option( 'wc_square_sync_auto_recovered_at', 0 );
+		if ( ! $recovered_at ) {
+			return;
+		}
+
+		// Once a sync has completed successfully after the recovery, the notice is no longer relevant.
+		$last_synced_at = (int) wc_square()->get_sync_handler()->get_last_synced_at();
+		if ( $last_synced_at >= $recovered_at ) {
+			delete_option( 'wc_square_sync_auto_recovered_at' );
+			return;
+		}
+		?>
+		<div class="notice notice-warning is-dismissible">
+			<p>
+				<?php
+				echo wp_kses(
+					sprintf(
+						/* translators: Placeholders: %1$s - <a> tag, %2$s - </a> tag */
+						__( 'A Square sync job was interrupted and has been automatically restarted. Please verify your products are up to date, or manually trigger a sync from the %1$supdate page%2$s.', 'woocommerce-square' ),
+						'<a href="' . esc_url( admin_url( 'admin.php?page=wc-settings&tab=square&section=update' ) ) . '">',
 						'</a>'
 					),
 					array(
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -497,12 +497,12 @@ class Admin {
 			return;
 		}
 
-		// Once a sync has completed successfully after the recovery, the notice is no longer relevant.
-		$last_synced_at = (int) wc_square()->get_sync_handler()->get_last_synced_at();
-		if ( $last_synced_at >= $recovered_at ) {
-			delete_option( 'wc_square_sync_auto_recovered_at' );
-			return;
-		}
+		// The flag itself is the only signal used here. It is cleared when a sync the merchant started
+		// begins, and again when such a job completes, so there is nothing left to infer.
+		//
+		// It deliberately does NOT compare against the last synced timestamp: interval polling advances
+		// that every few minutes, so any comparison against it would dismiss this notice long before
+		// the merchant saw it, which is the same reason polls are excluded from both clearing paths.
 		?>
 		<div class="notice notice-warning is-dismissible">
 			<p>
--- a/includes/Admin.php
+++ b/includes/Admin.php
@@ -475,7 +475,7 @@ class Admin {
 	 * Maybe show a notice that a stalled sync job was automatically recovered.
 	 *
 	 * Shown when auto-recovery has fired and no successful sync has completed since, so the
-	 * merchant knows the previous sync was interrupted and restarted and can verify their products.
+	 * merchant knows a stuck sync was stopped and can start a new one.
 	 * Clears itself once a sync completes successfully after the recovery.
 	 *
 	 * @since x.x.x
@@ -510,7 +510,7 @@ class Admin {
 				echo wp_kses(
 					sprintf(
 						/* translators: Placeholders: %1$s - <a> tag, %2$s - </a> tag */
-						__( 'A Square sync job was interrupted and has been automatically restarted. Please verify your products are up to date, or manually trigger a sync from the %1$supdate page%2$s.', 'woocommerce-square' ),
+						__( 'A stuck Square sync job was detected and automatically stopped. Product data may be out of date, please start a new sync from the %1$supdate page%2$s.', 'woocommerce-square' ),
 						'<a href="' . esc_url( admin_url( 'admin.php?page=wc-settings&tab=square&section=update' ) ) . '">',
 						'</a>'
 					),
--- a/includes/Handlers/Background_Job.php
+++ b/includes/Handlers/Background_Job.php
@@ -168,10 +168,12 @@ class Background_Job extends Background_Job_Handler {
 			$job->status                = 'processing';
 			$job->started_processing_at = current_time( 'mysql' );
 
-			$job = $this->update_job( $job );
+			$this->update_job( $job );
 
-			// The row can be gone if the job was cleared concurrently (e.g. the Clear Square Sync
-			// tool), so guard before dereferencing the job below.
+			// Re-read the row: update_job() returns the supplied object even when the underlying
+			// option no longer exists (e.g. the Clear Square Sync tool removed it concurrently),
+			// so only a fresh read can prove the job is still there.
+			$job = $this->get_job( $job->id );
 			if ( ! $job ) {
 				return;
 			}
--- a/includes/Handlers/Background_Job.php
+++ b/includes/Handlers/Background_Job.php
@@ -425,15 +425,34 @@ class Background_Job extends Background_Job_Handler {
 		// A runner action is still queued: the queue may be paused, not dead (low traffic sites can
 		// go quiet long enough for the threshold to pass, then resume on the visit that triggered
 		// this very check). Give the queue one grace window to make progress; recover only if the
-		// job is still stalled with the same queued action after the grace period.
+		// job is still stalled with the same queued action after that window.
 		if ( function_exists( 'as_next_scheduled_action' ) && false !== as_next_scheduled_action( 'wc_square_job_runner' ) ) {
+
+			/**
+			 * Filters how long a stalled sync job is given to resume when a job runner action is
+			 * still queued, before it is treated as dead.
+			 *
+			 * A queued action means the queue may simply be paused rather than broken, and the
+			 * request that runs this check usually gives Action Scheduler its chance to run, so this
+			 * only needs to be long enough for that to happen. It is deliberately shorter than the
+			 * stall threshold: with both at their defaults a paused queue is left alone for 15
+			 * minutes and a genuinely dead one is failed after 20, not 30.
+			 *
+			 * @since x.x.x
+			 *
+			 * @param int $grace_period grace period in seconds (default a third of the stall threshold)
+			 */
+			$grace_period = max( MINUTE_IN_SECONDS, (int) apply_filters( 'wc_square_stuck_job_grace_period', (int) round( $threshold / 3 ) ) );
+
 			$grace_started = (int) get_option( 'wc_square_recovery_grace_at', 0 );
+
 			if ( ! $grace_started ) {
 				update_option( 'wc_square_recovery_grace_at', time(), false );
-				wc_square()->log( 'Stalled sync has a queued runner action; allowing a grace window before auto-failing.' );
+				wc_square()->log( sprintf( 'Stalled sync has a queued runner action; allowing %d seconds for the queue to resume before auto-failing.', $grace_period ) );
 				return;
 			}
-			if ( ( time() - $grace_started ) < $threshold ) {
+
+			if ( ( time() - $grace_started ) < $grace_period ) {
 				return;
 			}
 		}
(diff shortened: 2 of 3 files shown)


includes/Handlers/Background_Job.php
```diff
@@ -157,12 +162,24 @@ public function process_job( $job, $items_per_batch = null ) {
 			return;
 		}
 
+		// Heartbeat: record activity on every processed step so a genuinely long running sync is
+		// never mistaken for a stalled one. started_processing_at (below) is stamped only once, so
+		// it cannot tell a slow large-catalog sync apart from a stuck job; last_activity_at can.
+		$job->last_activity_at = time();
```
iamdharmesh: This will be updated for every retry. So, for issues like action scheduler timeout, it won't be ever stuck, the sync will remain stuck, and it won't be ever cleared. I think better to move this after processing done and  just before the return $job. what do you think?

includes/Handlers/Background_Job.php
```diff
@@ -304,6 +324,166 @@ public function run_clear_background_jobs() {
...
+		if ( ! $is_stalled( $job ) ) {
+			return;
+		}
+
+		// Recover: fail the stalled job, release the lock, and clear the action cascade.
+		$this->fail_job( $job, __( 'Sync job stalled and was automatically recovered.', 'woocommerce-square' ) );
+		$this->unlock_process();
+
+		if ( function_exists( 'as_unschedule_all_actions' ) ) {
+			as_unschedule_all_actions( 'wc_square_job_runner' );
```
iamdharmesh: I think we should keep this action scheduled to process any other sync job in the queue. what do you think?

includes/Admin.php
```diff
@@ -467,4 +470,59 @@ public function maybe_show_sync_status_notice() {
...
+			return;
+		}
+		?>
+		<div class="notice notice-warning is-dismissible">
+			<p>
+				<?php
+				echo wp_kses(
+					sprintf(
+						/* translators: Placeholders: %1$s - <a> tag, %2$s - </a> tag */
+						__( 'A Square sync job was interrupted and has been automatically restarted. Please verify your products are up to date, or manually trigger a sync from the %1$supdate page%2$s.', 'woocommerce-square' ),
```
iamdharmesh: I am not sure if we are restarting the sync. are we restarting sync? I believe we only stop/fail the stucked job.

iamdharmesh (changes requested):
Thanks for the PR @faisal-alvi. Shared some initial feedback.

includes/Handlers/Background_Job.php
```diff
@@ -304,6 +329,164 @@ public function run_clear_background_jobs() {
...
+		// Recover: fail the stalled job and release the lock. Scheduled job_runner actions are
+		// deliberately left in place so any other sync job waiting in the queue keeps processing;
+		// the failed job no longer comes back from get_job(), so the next run picks up the rest.
+		$this->fail_job( $job, __( 'Sync job stalled and was automatically stopped.', 'woocommerce-square' ) );
+		$this->unlock_process();
+
+		// Recorded so the admin notice can tell the merchant a stalled sync was stopped.
+		update_option( 'wc_square_sync_auto_recovered_at', time() );
+
+		wc_square()->log( 'Auto-recovered a stalled sync job (' . ( isset( $job->id ) ? $job->id : 'unknown' ) . '). The job was marked failed and the queue lock released.' );
```
iamdharmesh: Nit:
Maybe we should say it Auto-failed? to make it clear as it marked as failed instead of any recovery for complete the pending sync job. what do you think?

includes/Handlers/Background_Job.php
```diff
@@ -314,6 +497,16 @@ public function run_clear_background_jobs() {
 	 */
 	public function handle_sync_healthcheck() {
 
+		// Auto-recover a stalled sync first, on purpose. A job stuck in "processing" (timeout, fatal,
+		// worker kill) or a cascade of failing wc_square_job_runner actions keeps the queue
+		// non-empty, so the as_has_scheduled_action() guard below would otherwise never let the sync
+		// restart. Running recovery before the early returns is what breaks that deadlock.
+		$this->maybe_recover_stuck_sync();
```
iamdharmesh, line 730: Just thinking out loud that, for sites with very low visitors, it may be possible the sync is paused due to the in-activity and when we have some visits after some time when 15 mins have already passed, how do we prevent that false positive signal from failing the actual running sync? Just wanted to make sure that we are not failing the proper job due to false positive signal.

iamdharmesh (commented):
Thanks for the changes @faisal-alvi. This looks good overall. There are some minor comments from Copilot which are worth checking.

Also, have you tried to reproduce the bug in simulating the actual setup (not by setting up options directly)? I have tried adding memory and time-out errors, and every time it ended up with sync failed without the sync stuck. So, recovery logic never ran for me. Could you please provide me with steps on how you verified this? Thanks!

includes/Handlers/Background_Job.php
faisal-alvi: Your concern is valid. In `maybe_recover_stuck_sync()`, before failing a stale job, I added a check to see if a `wc_square_job_runner` action is still queued in Action Scheduler.

includes/Handlers/Background_Job.php
```diff
@@ -197,6 +221,9 @@ public function process_job( $job, $items_per_batch = null ) {
 	 */
 	public function job_complete( $job ) {
 
+		// A sync completed successfully, so clear any pending auto-recovery notice.
+		delete_option( 'wc_square_sync_auto_recovered_at' );
```
iamdharmesh: I think we should delete this when the new sync gets started. For the big sync, this notice will stay for a while and create confusion if we remove it after the sync is complete. what do you think?

includes/Handlers/Background_Job.php
```diff
@@ -304,6 +331,202 @@ public function run_clear_background_jobs() {
...
+		// this very check). Give the queue one grace window to make progress; recover only if the
+		// job is still stalled with the same queued action after the grace period.
+		if ( function_exists( 'as_next_scheduled_action' ) && false !== as_next_scheduled_action( 'wc_square_job_runner' ) ) {
+			$grace_started = (int) get_option( 'wc_square_recovery_grace_at', 0 );
+			if ( ! $grace_started ) {
+				update_option( 'wc_square_recovery_grace_at', time(), false );
+				wc_square()->log( 'Stalled sync has a queued runner action; allowing a grace window before auto-failing.' );
+				return;
+			}
+			if ( ( time() - $grace_started ) < $threshold ) {
```
iamdharmesh: Que: As we are checking for grace till the threshold as well. The Stalled sync will be failed after 2 X Threshold time, right?

includes/Handlers/Background_Job.php
iamdharmesh: Auto-failed a stalled sync job

iamdharmesh (commented):
Thanks @faisal-alvi. This looks good now. I just added few questions. Could you please help to check once?

includes/Handlers/Background_Job.php
faisal-alvi: Agreed, the flag is now cleared the moment a job starts processing, so the notice disappears as soon as the merchant starts a new sync. I left the clear in `job_complete()` as a fallback for a job that was already running when the release is installed.

includes/Handlers/Background_Job.php
faisal-alvi: Yes, that was correct for the version you read: 15 minutes to notice plus another 15 of grace, so 30 at the defaults. Too slow, so the grace is now a third of the threshold and has its own filter (`wc_square_stuck_job_grace_period`).

Now, at the defaults:

- No runner action queued (the case in this ticket): failed after 15 minutes, no grace.
- Runner action still queued (your low traffic case): 20 minutes, because the queue might just be paused.

includes/Handlers/Background_Job.php
faisal-alvi: Done, the log line now reads "Auto-failed a stalled sync job..."

includes/Handlers/Background_Job.php
faisal-alvi: One correction, after adding a few more commits. The flag is cleared when a sync the merchant started begins processing, not when any job does. Interval polls are deliberately excluded, since they start on their own every few minutes and would dismiss the notice before anyone had a chance to read it. `job_complete()` still clears it as a fallback for a job that was already running when the release is installed.

includes/Handlers/Background_Job.php
```diff
@@ -304,6 +370,345 @@ public function run_clear_background_jobs() {
...
+			return false;
+		}
+
+		// A live worker still holds the process lock: the job is progressing, not stuck. Leave it.
+		if ( $this->is_process_running() ) {
+			return false;
+		}
+
+		// A runner action was touched recently, so a worker is still on it. The process lock only
+		// lasts 60 seconds, so a single long step outlives it and would otherwise look abandoned.
```
iamdharmesh, line 491: Nit: I think we should increase the process lock to 300 secs, to match with Action scheduler timeout. What do you think? Not necessarily on this PR but on follow-up?

includes/Handlers/Background_Job.php
```diff
@@ -197,6 +255,10 @@ public function process_job( $job, $items_per_batch = null ) {
 	 */
 	public function job_complete( $job ) {
 
+		// Normally cleared when the sync started; repeated here for a job that was already running
+		// when this release was installed.
+		delete_option( 'wc_square_sync_auto_recovered_at' );
```
iamdharmesh: Nit: Should we prevent deleting this for poll sync, as we are doing in the job start?

iamdharmesh (approved):
Thanks @faisal-alvi. This looks good to me. Added few nit comment to check but pre-approving this.

includes/Handlers/Background_Job.php
faisal-alvi: Agreed, and I think you are right that it belongs in a follow-up rather than here.

iamdharmesh: Thanks @faisal-alvi. Could you please help with create the Linear issue for the same?

qasumitbagthariya:
### QA Update  ✅ 
-------------

I have verified this PR in the `fix/SQUARE-291` branch, which has been fixed and is functioning as intended.

- Verified stuck Square sync recovery, admin notices, logging, manual sync recovery, and regression scenarios. Confirmed active and paused queues are not incorrectly marked as stalled.

- Reloading admin after injecting a stuck job clears the "Syncing now." state, re-enables Sync now, and adds a Failed record reading exactly "A sync stopped responding and was stopped automatically." ✅

- The dismissible notice shows on any admin screen (checked Dashboard too), matches the expected copy. ✅

- Notice disappears the moment a manual sync starts, not when it finishes. ✅

- Log records "Auto-failed a stalled sync job (STUCKQA)." exactly as expected. ✅

- Grace-window regression: a job with a queued-but-not-due runner action stays "processing" and logs "allowing 60 seconds." Only fails once that window passes. ✅


image

image

🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/SQUARE-291

</Details>


🔁 Next Step  -  Ready for UAT

Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

qasumitbagthariya:
@faisal-alvi 

E2E tests are failing. Could you please check from your end?

faisal-alvi:
@qasumitbagthariya the E2E is now passing. The issue was the timing: I waited until the Actions queue was empty before re-running, so nothing else was mutating the shared Square sandbox.

PR #529, opened 2026-08-04
Stop Square inventory pulls from zeroing WooCommerce stock and honor the per product sync setting

faisal-alvi:
### All Submissions:

* [x] Will this change require new documentation or changes to existing documentation?

---

The short version: _treat **"Square says zero"** as a question, not an answer, and go find out whether that zero has a history behind it._

Merchants have repeatedly reported WooCommerce stock being wiped to zero by the Square inventory sync, including stores where WooCommerce is the system of record. The root cause is that Square reports a tracked item that has never had a count recorded as "0 in stock", which is byte for byte identical to a genuine sellout. Every pull path treated that ambiguous zero as real data, wrote it to the product and switched stock management on. A single shopper adding an item to their basket could trigger it.

This change introduces one shared write policy for every Square to WooCommerce inventory write:

1. A positive count is always trusted and written, keeping the existing behavior of mirroring Square tracking into stock management.
2. A zero count never changes the product's stock management setting. For a stock managed product, the zero is written only when Square's inventory change history proves a real count was ever recorded for that item. A never counted item has an empty history, so its phantom zero is skipped and logged. No new meta is introduced; Square's own history is the source of truth.
3. A product that does not manage stock is never given a quantity by a zero or a negative count. It is marked out of stock only when the zero is proven to be a real sellout, and a zero is never a reason to force it back in stock, so an unproven zero leaves such a product exactly as the merchant left it. A positive count still mirrors Square as it does on trunk, writing the quantity and switching stock management on; whether that should instead defer to a merchant who deliberately turned stock management off is a separate policy question, deferred with the tracking off follow up.
4. A variation that inherits the parent product's shared stock pool is skipped for every count, positive or zero, since a per variation count must not alter stock shared with its sibling variations or convert the variation to its own stock management.
5. A negative count needs no history check, because an item that was never counted reads exactly zero, so a negative can only come from real movement. It is written through rather than flattened, which keeps the oversell depth a store using backorders relies on.
6. A response that contains no usable IN_STOCK count for an item means "no information" and never a zero. Non IN_STOCK states such as purchase order stages are ignored as well, which also resolves the purchase order zeroing report.

On the push side, the count builder now decides from the product's state: a managed product with an unresolved quantity is skipped and logged instead of pushing a fabricated zero, a product without stock management that is out of stock pushes an explicit zero so Square marks it sold out, and one that is in stock pushes nothing.

The automatic pulls (hourly polling, the manual sync pull, and the add to basket stock refresh) now also honor the per product "Sync with Square" setting, so unticking it or using Unlink genuinely stops Square from updating that product. Two paths stay deliberately ungated because the merchant asked for them directly: the "Fetch stock from Square" button, and the product import, which additionally cannot use that flag because it is only set for newly imported products, so gating it would silently skip inventory for products the import matched and updated.

If Square genuinely sold out (say an in person sale) and our verification call fails, this path leaves WooCommerce holding the older quantity rather than writing the 0. Stock stays stale until a later trigger or the next poll writes it. That is a deliberate trade: a brief oversell window that already existed because of the poll interval, versus zeroing live products under shoppers.

**Cost of the history check:** it runs only for items reporting zero, so a store with normal stock levels pays nothing. Ids are asked about in batches of 100, and any id still unresolved after the first page is re-asked as a group with the already answered ids removed, so a batch of never counted items is normally settled in one extra request and a single high activity item cannot drag a whole batch through its entire history just to prove a quiet item empty. Only if that group pass stops making progress does it fall back to one request per id, and the number of group passes is capped.

**When Square cannot be asked:** the history lookup is the one new dependency, so its failure is handled explicitly rather than guessed at. A failed lookup is reported distinctly from a verified empty history, and nothing is written on it. A step holds its progress and runs again a bounded number of times, and it never throws, because the job runner fails a job on any exception other than a rate limit, which would abort an entire import or manual sync for one transient error.

Recovery after those attempts differs by path, because the paths read Square differently, and there is deliberately no separate retry mechanism for it. Interval polling reads counts by `updated_after`, so every object whose zero it could not verify was inside that window by definition: when the attempts run out it keeps the window instead of advancing it, and the next poll reads the same window and finds those exact objects again. The manual sync and the import read counts by id rather than by window, so their recovery is to run the sync again, which re-checks exactly those objects. Both paths record one merchant visible alert rather than one per batch, telling the merchant to run the sync again once Square is responding normally. The dedupe is a job attribute plus a six hour transient (`wc_square_zero_verification_alerted`), so a store polling hourly gets at most one alert every six hours rather than one per window. Nothing anywhere writes a zero it could not confirm.

**Square API limitation worth knowing:** the sold out flag on a Square location override is effectively read only. Square only marks an item sold out through inventory tracking with a count of zero, and an override carrying sold out without tracking is silently dropped. That is why the sync must keep enabling tracking for out of stock products that do not manage stock. This change therefore fixes the zeroing on the pull side and also releases the sold out state when such a product comes back in stock, which previously stayed sold out in Square forever.

**A second Square API behaviour worth knowing:** the inventory endpoints honor the catalog object id filter only while at least one of the supplied ids still exists in Square. If none of them exists, Square silently answers as though no filter had been sent, returning the whole location's history or counts. A non empty response is therefore not proof that the id you asked about has any history. The verification now requires a response to name the id it asked about, and reads an unfiltered answer as a conclusive "none of these ids exist in Square", which also means none of them can have sold out. To be precise about the impact, since an earlier revision of this description overstated it: I could not reach a stock write through that path, because the refresh iterates the catalog objects Square returns and an id Square no longer knows returns none. So this is defensive correctness, plus saving up to five plus N wasted requests for a chunk of dead ids, rather than a demonstrated data loss fix.

**Deliberate behavior changes to call out in release notes:**
* When WooCommerce is the system of record, the manual sync still pulls inventory, but it no longer reads back the counts the same job just pushed. Those catalog objects are excluded by id, because Square's inventory is eventually consistent and a count that has not propagated yet answers zero, which would overwrite the stock pushed moments earlier. Products that already existed in Square are still pulled, which is what keeps sales made on other channels reflected and is the only thing that syncs their inventory during a manual sync.
* The documented "Square updates WooCommerce stock hourly regardless of system of record" behavior now respects the per product sync setting and the write policy above.

**Holding the read window has no cap, deliberately:** while verification is unavailable the poll keeps its inventory window rather than advancing, so during a sustained outage the window grows with every poll and the reads get larger, until Square answers again and it advances, at which point the cost disappears. An earlier revision capped consecutive holds and advanced anyway, which was only safe while a separate retry list re-checked the skipped objects by id. That list is gone, so a cap would now advance past a window nothing revisits, which is the silent loss this ticket exists to prevent.

**Opting a product out, and mirroring a tracking change:** unticking "Sync with Square", or using Unlink, stops Square touching a product at all, and this change makes that effective on every automatic pull, since all four pull paths now check it. Separately, "Manage stock" is editable again on a synced product, so a merchant can mirror turning inventory tracking off in Square: with tracking off there, the sync reflects availability only and leaves the flag alone. Verified live, the untick survives a poll with the quantity cleared and the product left in stock. What it is not is a way to diverge from Square: if Square is still tracking the item, a positive count mirrors as it always has and switches stock management back on, also verified live. Whether that divergence should be permitted is the deferred question in point 3.

**What the test numbers in these threads mean:** the repository's PHPUnit suite is 54 tests, 23 of them added here, covering the zero discriminator, the write policy and the retry policy, and each was checked by reverting the corresponding part of the fix and confirming the test fails. Figures quoted in review threads come from scripted runs against a live Square sandbox rather than unit tests: a seven scenario behaviour suite (phantom, genuine sellout, unmanaged product, pooled variation, unticked product, proven unmanaged sellout, negative count), a sixteen case discriminator suite driven by scripted API pages, and end to end runs through the real job runner including an injected Square outage at the SDK boundary.

**Known limits and how this was verified:** all of the behavior above was exercised against a live Square sandbox through the real sync paths, by walking the manual scenarios below and by scripted runs of the same paths, including an injected Square outage at the API boundary to cover the failure branches. It is also covered by unit tests now, described below. The Square system of record import path is covered by the shared policy but has not been exercised end to end at scale in this round of dev-testing, purchase order state behavior is inferred from Square documentation since a Square Retail account is required to generate those states, and the remaining place a zero can be delayed is a manual sync or an import whose history lookups stay unreadable: those proceed with the alert rather than failing, and re-running the sync is what re-checks them.

**Unit tests:** 24 tests were added under `tests/php/Unit/Sync`, runnable with `composer test-unit`. They cover the discriminator (an id is only proven by a response that names it, a failed lookup is distinct from a verified empty history, the group narrowing and its per id fallback, the request chunking), the write policy (positive counts, verified and unverified zeros for managed and unmanaged products, pooled variations skipped for every count under WooCommerce SOR and converted by a positive count under Square SOR, negatives passed through), and the retry policy (the step is held for a bounded number of attempts, exhaustion is reported so a watermark can be held, a success clears it, and the merchant is alerted at most once every six hours). Each of them was checked by reverting the corresponding part of the fix and confirming the test fails, so they are regression tests rather than descriptions. Note that the repository has no PHPUnit job in CI, so these run locally.

**One change outside the sync code, and why it is here:** the plugin's main file assigned the return of `require_once vendor/autoload.php` and called `addPsr4()` on it. That return is the Composer ClassLoader on first load and simply `true` afterwards, so as soon as anything had already loaded the same autoloader, that line fataled. PHPUnit does exactly that through phpunit-polyfills, which is why WooCommerce could not be loaded into the test harness at all and why 28 of the repository's existing tests were skipping. The call is now guarded, with a direct namespace registration as a fallback so registration still happens rather than being silently skipped. Behaviour on the normal WordPress path is unchanged, since the first load still returns the ClassLoader, and it was verified on a live connected site: admin, both Square settings screens and the products list load without fatals and the plugin resolves its own classes. With it in place the suite goes from 42 tests with 28 skipped to 54 with 1 skipped, and the write policy and retry policy tests above become possible, since those need real WooCommerce product objects.

Closes https://linear.app/a8c/issue/SQUARE-145/square-zero-ing-out-inventory
Closes https://linear.app/a8c/issue/SQUARE-359/interval-inventory-sync-overwrites-products-whose-sync-with-square-is

> [!Note]
> This is a big change; I'd recommend doing a random QA before proceeding with the following instructions. Just make sure the sync process works as expected, especially when Woo is SOR. Try with different QTY and Manage Stock, Inventory Sync On/Off combinations and make sure all work as expected.

Setup once: connect the site to a Square sandbox, set System of Record to WooCommerce, and keep the Square sandbox Dashboard open in another tab. Every scenario below is done through the normal admin and shop screens, no tools involved. "Run a sync" means WooCommerce > Settings > Square > Update > Sync now, and "trigger a stock refresh" means adding the product to the basket on the shop front end, which is the automatic path a shopper takes.

**Scenario A. The reported bug: stock wiped by a meaningless zero**

1. In WooCommerce > Settings > Square, turn "Sync inventory" OFF.
1. Create a simple product with a unique SKU, tick "Manage stock", set the quantity to 20, tick "Sync with Square" in the Inventory tab, and update.
1. Run a sync and wait for it to finish. The item now exists in Square but has never had a stock count recorded there.
1. Turn "Sync inventory" ON and save.
1. Trigger a stock refresh for that product.
1. Open the product's Inventory tab. On trunk the stock data is destroyed even though nobody bought anything: either the quantity becomes 0 and the product goes out of stock, or "Manage stock" is switched off and the quantity disappears, leaving the product sellable with no stock control. On this branch nothing changes: "Manage stock" stays ticked and the quantity stays 20.

**Scenario B. A genuine Square sellout must still reach WooCommerce**

1. With "Sync inventory" ON, create a managed product with quantity 5, tick "Sync with Square", run a sync, and confirm in the Square Dashboard that the item shows 5 in stock.
1. In the Square Dashboard set that item's stock to 0.
1. Trigger a stock refresh, then check the product: the quantity is 0 and the product is out of stock. This must behave the same on trunk and on this branch.

**Scenario C. An unproven zero must not touch a product that does not manage stock**

1. Turn "Sync inventory" OFF. Create a product with "Manage stock" unticked and stock status "In stock", tick "Sync with Square", and run a sync.
1. Turn "Sync inventory" ON, then trigger a stock refresh.
1. Reopen the product. On this branch nothing at all has changed: "Manage stock" is still unticked, no quantity has appeared, and the product is still in stock. Square never proved a sellout, so it does not get to stop the product selling.

**Scenario D. A proven sellout does stop such a product selling**

1. Take the product from scenario C. In the Square Dashboard, enable inventory tracking for that item and set its stock to 5, then set it to 0.
1. Trigger a stock refresh and reopen the product: it now shows out of stock, "Manage stock" is still unticked, and no quantity has been written.

**Scenario E. Turning tracking off in Square must not switch off Manage stock in WooCommerce**

1. Take any synced product with "Manage stock" ticked and a quantity.
1. In the Square Dashboard, turn inventory tracking OFF for that item.
1. Trigger a stock refresh and reopen the product: "Manage stock" is still ticked and the quantity is unchanged. Stock management is the merchant's setting, so the pull reflects availability but never edits that checkbox. On trunk this switched Manage stock off (on the customer database export from SQUARE-364, a single poll did that to 4,903 products).

**Scenario F. Unticking "Sync with Square" must really exclude the product**

1. Take a synced managed product with stock, untick "Sync with Square" in its Inventory tab, and update.
1. In the Square Dashboard set that item's stock to 0.
1. Trigger a stock refresh: the WooCommerce quantity is unchanged. Square no longer touches this product.

**Scenario G. A variation sharing the parent product's stock must be protected**

1. Create a variable product, tick "Manage stock" on the parent, set the parent quantity to 50, and add a variation that leaves "Manage stock" unticked so it uses the parent's stock.
1. Tick "Sync with Square", run a sync, and confirm the variation exists in Square.
1. In the Square Dashboard set that variation's stock to 0.
1. Trigger a stock refresh and reopen the product: the parent quantity is still 50 and the variation still uses the parent's stock. One variation's Square count must not spend stock shared with its siblings.

**Scenario H. Push side: a product without stock management that goes out of stock**

1. Create a product with "Manage stock" unticked and stock status "Out of stock", tick "Sync with Square", run a sync.
1. In the Square Dashboard the item shows as sold out.
1. Set the same product back to "In stock" in WooCommerce and run a sync again: the item is no longer sold out in Square. On trunk it stayed sold out.

**Scenario I. An order placed after inventory sync is enabled**

1. Using scenario A's product, place an order for 1 through the shop.
1. WooCommerce reduces the quantity, and the Square Dashboard records the sale for that item.
1. Run a sync: Square is set back to WooCommerce's absolute quantity, so both sides agree. Note that immediately after the order the Square figure can be lower than WooCommerce's (it may even be negative for an item that never had a count), and the sync reconciles it.

**Scenario J. Square as the system of record: import**

1. Switch System of Record to Square, and in the import options allow existing products to be updated.
1. Run an import and check a product that already existed in WooCommerce and was matched by the import: its stock reflects Square. On this branch inventory is applied to matched products too, not only to newly created ones.

**Optional, for two shapes that are impractical by hand**

A single file QA tool can create the fixtures for the remaining shapes, which need a Square item to be oversold or need a specific pooled setup: https://gist.github.com/faisal-alvi/db2d5f0cffa0cf2ebd336134898e3733 (drop it in wp-content/mu-plugins/, then Tools > Square Stock Zeroing QA, and use button 8 to clean up). The one worth running is test 7, an item oversold at the till: Square reports minus three and WooCommerce must record minus three rather than flattening it to zero, which is what a store using backorders relies on.

Behaviour that only appears when the Square API itself is failing (the bounded retries, the held inventory sync window, and the guarantee that a failing history lookup never fails an import or a manual sync) cannot be produced from the admin, so it was verified during development by injecting a Square outage at the API boundary. It is described in the section above rather than left implicit.

### Review pass 2026-08-25

Three behaviours were changed after review feedback, all of them narrowing this PR to the mode the bug is actually in.

**`manage_stock` is mirrored on every pull path, and only when Square is the system of record.** Trunk switched stock management off in both modes, which is how 4,903 products on the reported store were left sellable with no stock control. This branch keeps that behaviour for Square SOR, where Square owns the setting, and removes it for WooCommerce SOR, where it is merchant intent. It was previously applied at two of the six places that write stock from Square; it is now applied at all six: `Product::update_stock_from_square()`, `Product::update_products_stock_from_square()` (the add to cart refresh), both interval poll steps, `Manual_Synchronization::pull_inventory()` and `Product_Import::import_inventory()`.

**A variation inheriting its parent's pooled stock follows the same rule.** Under WooCommerce SOR a per variation Square count is skipped, because writing it is either invisible (reads come from the parent) or would convert the variation to its own stock management as a side effect of an inventory read. Under Square SOR a positive count is applied and the variation takes over its own stock, which is what the plugin did before this changeset. A zero or negative count is skipped in both modes, since writing one into a pool would move stock shared with sibling variations on one variation's reading. Both outcomes are logged.

**The interval poll no longer resaves products it has nothing to change.** Its skip check compared Square's tracking flag against `manage_stock`, which only settles once something makes them agree. Trunk made them agree by switching stock management off; this branch does not do that under WooCommerce SOR, so a product tracked off in Square with Manage stock on in WooCommerce failed that check on every run and was saved every hour indefinitely. On the reported store's shape that is roughly 4,900 saves per poll. The tracking comparison now applies only under Square SOR, where the poll can still change the setting; under WooCommerce SOR availability is the only thing the poll writes, so it is the only thing compared.

Known and deliberately left as follow ups: the order path performs a Square inventory history lookup during checkout to establish a baseline for an item Square never counted, which should become a cached per product flag; and a push releases inventory tracking on any pre-existing location override at the configured location, including one that exists only for a location specific price.

### Changelog entry

> Fix - Stop Square inventory pulls from writing zero stock for items Square has no real count for, never change a product's stock management setting on a zero, and honor the per product Sync with Square setting on all automatic pulls.

### DIFF
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -457,6 +457,23 @@ class Product {
 	 * @throws \Exception
 	 */
 	public static function update_products_stock_from_square( $product_ids ) {
+
+		// Respect the per-product "Sync with Square" setting: this method backs automatic pulls
+		// (e.g. the add to cart inventory refresh), and a product the merchant unticked or
+		// unlinked must not have its stock overwritten from Square. Filtering up front also
+		// avoids needless Square API calls for excluded products.
+		$product_ids = array_filter(
+			(array) $product_ids,
+			function ( $product_id ) {
+				$product = wc_get_product( $product_id );
+				return $product instanceof \WC_Product && self::is_synced_with_square( $product );
+			}
+		);
+
+		if ( empty( $product_ids ) ) {
+			return;
+		}
+
 		$products_map = self::get_square_meta( $product_ids, 'square_item_variation_id' );
 		$square_ids   = array_keys( $products_map );
 
@@ -478,9 +495,19 @@ class Product {
 			$inventory_hash     = Helper::get_catalog_objects_inventory_stats( $square_ids );
 			$inventory_tracking = Helper::get_catalog_inventory_tracking( $response->get_data()->getObjects() );
 
+			// Verify zero counts against Square's inventory change history: a never-counted item
+			// reports IN_STOCK 0 exactly like a real sellout, and only real zeros may be written
+			// (SQUARE-145).
+			$zero_object_ids = array();
+			foreach ( $inventory_hash as $object_id => $object_quantity ) {
+				if ( 0.0 === (float) $object_quantity ) {
+					$zero_object_ids[] = $object_id;
+				}
+			}
+			$verified_zero_ids = Helper::get_catalog_objects_with_inventory_history( $zero_object_ids );
+
 			foreach ( $response->get_data()->getObjects() as $catalog_object ) {
 				$square_id               = $catalog_object->getId();
-				$stock                   = $inventory_hash[ $square_id ] ?? 0;
 				$inventory_tracking_data = $inventory_tracking[ $square_id ] ?? array();
 				$is_inventory_tracking   = $inventory_tracking_data['track_inventory'] ?? false;
 				$sold_out                = $inventory_tracking_data['sold_out'] ?? false;
@@ -492,12 +519,27 @@ class Product {
 					continue;
 				}
 
-				if ( $is_inventory_tracking ) {
-					$product->set_manage_stock( true );
-					$product->set_stock_quantity( $stock );
-				} else {
+				if ( $is_inventory_tracking && isset( $inventory_hash[ $square_id ] ) ) {
+
+					$changed = Helper::apply_square_inventory_count(
+						$product,
+						(float) $inventory_hash[ $square_id ],
+						(bool) $sold_out,
+						in_array( $square_id, $verified_zero_ids, true )
+					);
+
+					if ( ! $changed ) {
+						continue;
+					}
+				} elseif ( ! $is_inventory_tracking ) {
+
+					// Not tracked in Square: reflect availability only; manage_stock is merchant
+					// intent and is not changed here.
 					$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-					$product->set_manage_stock( false );
+				} else {
+
+					// Tracked but no IN_STOCK count returned: "no information", never a zero.
+					continue;
 				}
 
 				$product->save();
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -402,7 +402,8 @@ class Product {
 		}
 		$inventory_tracking = Helper::get_catalog_inventory_tracking( array( $result->get_data()->getObject() ) );
 
-		$stock = 0;
+		$stock        = 0;
+		$has_in_stock = false;
 
 		if ( $response->get_data() && $response->get_data()->getCounts() ) {
 
@@ -410,7 +411,8 @@ class Product {
 			foreach ( $response->get_data()->getCounts() as $count ) {
 
 				if ( 'IN_STOCK' === $count->getState() ) {
-					$stock += (float) $count->getQuantity();
+					$stock       += (float) $count->getQuantity();
+					$has_in_stock = true;
 				}
 			}
 		}
@@ -419,13 +421,23 @@ class Product {
 		$is_inventory_tracking   = $inventory_tracking_data['track_inventory'] ?? false;
 		$sold_out                = $inventory_tracking_data['sold_out'] ?? false;
 
-		if ( $is_inventory_tracking ) {
-			$product->set_manage_stock( true );
-			$product->set_stock_quantity( $stock );
-		} else {
+		if ( $is_inventory_tracking && $has_in_stock ) {
+
+			// A zero needs provenance: a never-counted item reports IN_STOCK 0 exactly like a real
+			// sellout, and only real zeros may be written (SQUARE-145).
+			$zero_verified = false;
+			if ( 0.0 === (float) $stock ) {
+				$zero_verified = in_array( $square_id, Helper::get_catalog_objects_with_inventory_history( array( $square_id ) ), true );
+			}
+
+			Helper::apply_square_inventory_count( $product, (float) $stock, (bool) $sold_out, $zero_verified );
+		} elseif ( ! $is_inventory_tracking ) {
+
+			// Not tracked in Square: reflect availability only; manage_stock is merchant intent
+			// and is not changed here.
 			$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-			$product->set_manage_stock( false );
 		}
+		// Tracked but no IN_STOCK count returned: "no information", never a zero.
 
 		if ( $save ) {
 			$product->save();
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -502,12 +502,7 @@ class Product {
 			// Verify zero counts against Square's inventory change history: a never-counted item
 			// reports IN_STOCK 0 exactly like a real sellout, and only real zeros may be written
 			// (SQUARE-145).
-			$zero_object_ids = array();
-			foreach ( $inventory_hash as $object_id => $object_quantity ) {
-				if ( 0.0 === (float) $object_quantity ) {
-					$zero_object_ids[] = $object_id;
-				}
-			}
+			$zero_object_ids   = Helper::zero_count_object_ids( $inventory_hash );
 			$verified_zero_ids = Helper::get_catalog_objects_with_inventory_history( $zero_object_ids );
 
 			if ( null === $verified_zero_ids ) {
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -427,7 +427,11 @@ class Product {
 			// sellout, and only real zeros may be written (SQUARE-145).
 			$zero_verified = false;
 			if ( 0.0 === (float) $stock ) {
-				$zero_verified = in_array( $square_id, Helper::get_catalog_objects_with_inventory_history( array( $square_id ) ), true );
+				$verified_ids  = Helper::get_catalog_objects_with_inventory_history( array( $square_id ) );
+				$zero_verified = is_array( $verified_ids ) && in_array( $square_id, $verified_ids, true );
+				if ( null === $verified_ids ) {
+					wc_square()->log( 'Could not verify the zero inventory count against Square history for product #' . $product->get_id() . '; stock left unchanged.' );
+				}
 			}
 
 			Helper::apply_square_inventory_count( $product, (float) $stock, (bool) $sold_out, $zero_verified );
@@ -505,6 +509,14 @@ class Product {
 				}
 			}
 			$verified_zero_ids = Helper::get_catalog_objects_with_inventory_history( $zero_object_ids );
+
+			if ( null === $verified_zero_ids ) {
+				// Verification unavailable: never throw on this path (it runs inside checkout via
+				// the add to cart refresh); treat every zero as unverified and let the next
+				// trigger retry naturally.
+				wc_square()->log( 'Could not verify zero inventory counts against Square history; zero stock writes skipped for this refresh.' );
+				$verified_zero_ids = array();
+			}
 
 			foreach ( $response->get_data()->getObjects() as $catalog_object ) {
 				$square_id               = $catalog_object->getId();
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -1568,21 +1568,54 @@ class Product {
 	 */
 	public static function get_inventory_change_physical_count_type( \WC_Product $product ) {
 
-		$inventory_change    = null;
 		$square_variation_id = self::get_square_item_variation_id( $product->get_id(), false );
-		if ( $square_variation_id ) {
 
-			$inventory_physical_count = new \Square\Models\InventoryPhysicalCount();
-			$inventory_physical_count->setCatalogObjectId( $square_variation_id );
-			$inventory_physical_count->setQuantity( '' . max( 0, $product->get_stock_quantity() ) );
-			$inventory_physical_count->setLocationId( wc_square()->get_settings_handler()->get_location_id() );
-			$inventory_physical_count->setState( 'IN_STOCK' );
-			$inventory_physical_count->setOccurredAt( gmdate( 'Y-m-d\TH:i:sP' ) );
-
-			$inventory_change = new \Square\Models\InventoryChange();
-			$inventory_change->setType( 'PHYSICAL_COUNT' );
-			$inventory_change->setPhysicalCount( $inventory_physical_count );
+		if ( ! $square_variation_id ) {
+			return null;
 		}
+
+		if ( $product->get_manage_stock() ) {
+
+			$quantity = $product->get_stock_quantity();
+
+			// An unresolved quantity must never be coerced to zero: pushing a fabricated 0 marks
+			// the item sold out in Square and poisons its inventory history (SQUARE-145).
+			if ( null === $quantity ) {
+				wc_square()->log(
+					sprintf(
+						'Skipped pushing inventory for product #%d: stock is managed but no quantity is set.',
+						$product->get_id()
+					)
+				);
+				return null;
+			}
+
+			// Square cannot hold negative counts; backordered stock is represented as zero.
+			$quantity = max( 0, $quantity );
+		} elseif ( 'outofstock' === $product->get_stock_status() ) {
+
+			// Not stock-managed but out of stock: push an explicit zero so the item is
+			// deterministically marked sold out in Square (sold_out is read-only and only becomes
+			// true through a tracked count of zero), including items that previously held a
+			// positive count.
+			$quantity = 0;
+		} else {
+
+			// Not stock-managed and in stock: there is no real number to send; the item is left
+			// untracked in Square, which is the sellable state.
+			return null;
+		}
+
+		$inventory_physical_count = new \Square\Models\InventoryPhysicalCount();
+		$inventory_physical_count->setCatalogObjectId( $square_variation_id );
+		$inventory_physical_count->setQuantity( '' . $quantity );
+		$inventory_physical_count->setLocationId( wc_square()->get_settings_handler()->get_location_id() );
+		$inventory_physical_count->setState( 'IN_STOCK' );
+		$inventory_physical_count->setOccurredAt( gmdate( 'Y-m-d\TH:i:sP' ) );
+
+		$inventory_change = new \Square\Models\InventoryChange();
+		$inventory_change->setType( 'PHYSICAL_COUNT' );
+		$inventory_change->setPhysicalCount( $inventory_physical_count );
 
 		return $inventory_change;
 	}
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -437,9 +437,15 @@ class Product {
 			Helper::apply_square_inventory_count( $product, (float) $stock, (bool) $sold_out, $zero_verified );
 		} elseif ( ! $is_inventory_tracking ) {
 
-			// Not tracked in Square: reflect availability only; manage_stock is merchant intent
-			// and is not changed here.
+			// Not tracked in Square: reflect availability. Whether stock management follows depends on
+			// who owns the setting. Under WooCommerce SOR it is merchant intent and is left alone, so a
+			// Square side toggle cannot silently invert the system of record. Under Square SOR the
+			// authority is reversed, so tracking off is mirrored onto the product.
 			$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
+
+			if ( wc_square()->get_settings_handler()->is_system_of_record_square() ) {
+				$product->set_manage_stock( false );
+			}
 		}
 		// Tracked but no IN_STOCK count returned: "no information", never a zero.
 
--- a/includes/Handlers/Product.php
+++ b/includes/Handlers/Product.php
@@ -546,9 +546,14 @@ class Product {
 					}
 				} elseif ( ! $is_inventory_tracking ) {
 
-					// Not tracked in Square: reflect availability only; manage_stock is merchant
-					// intent and is not changed here.
+					// Not tracked in Square: reflect availability. Stock management follows only when
+					// Square owns the setting; under WooCommerce SOR it is merchant intent and is
+					// left alone, so a Square side toggle cannot silently invert the system of record.
 					$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
+
+					if ( wc_square()->get_settings_handler()->is_system_of_record_square() ) {
+						$product->set_manage_stock( false );
+					}
 				} else {
 
 					// Tracked but no IN_STOCK count returned: "no information", never a zero.
--- a/includes/Handlers/Product/Woo_SOR.php
+++ b/includes/Handlers/Product/Woo_SOR.php
@@ -449,15 +449,21 @@ class Woo_SOR extends \WooCommerce\Square\Handlers\Product {
 					}
 				}
 
-				if ( ! $location_override ) {
+				// Out of stock needs an override, created if the item has none, because tracking is the
+				// only way to reach Square's sold out state. In stock only RELEASES tracking on an
+				// override that already exists: creating one purely to say "not tracked" would
+				// overwrite a merchant's own Square side tracking for an item WooCommerce does not
+				// manage, which is not ours to decide.
+				if ( ! $location_override && $is_out_of_stock ) {
 					$location_override = new \Square\Models\ItemVariationLocationOverrides();
 					$location_override->setLocationId( $configured_location );
 					$location_overrides[] = $location_override;
 				}
 
-				$location_override->setTrackInventory( $is_out_of_stock );
-
-				$variation_data->setLocationOverrides( $location_overrides );
+				if ( $location_override ) {
+					$location_override->setTrackInventory( $is_out_of_stock );
+					$variation_data->setLocationOverrides( $location_overrides );
+				}
 			}
 		}
 
--- a/includes/Handlers/Product/Woo_SOR.php
+++ b/includes/Handlers/Product/Woo_SOR.php
@@ -412,26 +412,52 @@ class Woo_SOR extends \WooCommerce\Square\Handlers\Product {
 			$location_overrides = $variation_data->getLocationOverrides();
 
 			/*
-			 * Only update track_inventory if it's not set.
+			 * Only update the base track_inventory if it's not set.
 			 * This will only update inventory tracking on new variations.
 			 * inventory tracking will remain the same for existing variations.
 			 */
 			if ( is_null( $track_inventory ) && is_null( $location_overrides ) ) {
 				$variation_data->setTrackInventory( $product->get_manage_stock() );
+			}
 
-				// If the product is not managing stock and is out of stock, set it as sold out.
-				if ( ! $product->get_manage_stock() && 'outofstock' === $product->get_stock_status() ) {
-					$configured_location = wc_square()->get_settings_handler()->get_location_id();
-					$location_override   = new \Square\Models\ItemVariationLocationOverrides();
-					$location_override->setLocationId( $configured_location );
-					// We need to set track_inventory to true to be able to set sold_out to true, without it will be ignored.
-					$location_override->setTrackInventory( true );
-					$location_override->setSoldOut( true );
-					$location_overrides = array( $location_override );
+			/*
+			 * For products that do not manage stock in WooCommerce, the configured location's
+			 * tracking override always follows the current WooCommerce availability:
+			 *
+			 * - out of stock: tracking on. Square's sold_out flag is read-only and only becomes
+			 *   true through a tracked count of zero, so tracking (plus the explicit zero count
+			 *   pushed by the inventory step) is the only API way to mark the item sold out.
+			 * - in stock: tracking off, which is the sellable state for an untracked item. This
+			 *   also clears the sold-out state when a product comes back in stock; previously the
+			 *   override was only ever written for new variations, so a product that went out of
+			 *   stock once stayed sold out in Square forever.
+			 *
+			 * The override is merged into any existing overrides so location price overrides are
+			 * preserved.
+			 */
+			if ( ! $product->get_manage_stock() ) {
+				$configured_location = wc_square()->get_settings_handler()->get_location_id();
+				$is_out_of_stock     = 'outofstock' === $product->get_stock_status();
 
-					// Set the location overrides.
-					$variation_data->setLocationOverrides( $location_overrides );
+				$location_overrides = is_array( $location_overrides ) ? $location_overrides : array();
+				$location_override  = null;
+
+				foreach ( $location_overrides as $existing_override ) {
+					if ( $existing_override->getLocationId() === $configured_location ) {
+						$location_override = $existing_override;
+						break;
+					}
 				}
+
+				if ( ! $location_override ) {
+					$location_override = new \Square\Models\ItemVariationLocationOverrides();
+					$location_override->setLocationId( $configured_location );
+					$location_overrides[] = $location_override;
+				}
+
+				$location_override->setTrackInventory( $is_out_of_stock );
+
+				$variation_data->setLocationOverrides( $location_overrides );
 			}
 		}
 
--- a/includes/Sync/Helper.php
+++ b/includes/Sync/Helper.php
@@ -349,21 +349,41 @@ class Helper {
 
 		$quantity = (float) $quantity;
 
-		// A variation inheriting stock management reports the string 'parent': its quantity and
-		// availability are governed by the parent's pooled stock, a quantity written to it is
-		// invisible (reads come from the parent) and a stock status write is overridden by the
-		// pool. A per-variation Square count, positive or zero, is not applicable data here; the
-		// pool is merchant intent and one variation's count must not alter stock shared by its
-		// siblings or convert the variation to its own management.
+		// A variation inheriting stock management reports the string 'parent': its quantity is
+		// governed by the parent's pooled stock, so a quantity written to it is invisible until the
+		// variation manages its own stock, and a stock status write is overridden by the pool.
+		//
+		// Who owns that decision depends on the system of record. Under WooCommerce SOR the pool is
+		// merchant intent, so a per-variation Square count is not applicable data and is skipped.
+		// Under Square SOR the authority is reversed: a positive count is applied and the variation
+		// takes over its own stock, which is what the plugin did before this changeset.
+		//
+		// A zero or negative count is skipped in both modes. Those are the counts that wiped stock
+		// (SQUARE-145), and writing one into a pool would move stock shared with sibling variations
+		// on the strength of a single variation's reading.
 		if ( 'parent' === $product->get_manage_stock() ) {
+
+			$square_is_system_of_record = wc_square()->get_settings_handler()->is_system_of_record_square();
+
+			if ( ! $square_is_system_of_record || $quantity <= 0 ) {
+				wc_square()->log(
+					sprintf(
+						'Skipped writing a stock quantity to variation #%1$d: its stock is managed by the parent product pool%2$s.',
+						$product->get_id(),
+						$square_is_system_of_record ? ' and the count was not positive' : ''
+					)
+				);
+
+				return false;
+			}
+
 			wc_square()->log(
 				sprintf(
-					'Skipped writing a stock quantity to variation #%d: its stock is managed by the parent product pool.',
-					$product->get_id()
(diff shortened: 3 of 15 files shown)


includes/Handlers/Product.php
```diff
@@ -419,13 +421,23 @@ public static function update_stock_from_square( \WC_Product $product, $save = t
...
+				$zero_verified = in_array( $square_id, Helper::get_catalog_objects_with_inventory_history( array( $square_id ) ), true );
+			}
+
+			Helper::apply_square_inventory_count( $product, (float) $stock, (bool) $sold_out, $zero_verified );
+		} elseif ( ! $is_inventory_tracking ) {
+
+			// Not tracked in Square: reflect availability only; manage_stock is merchant intent
+			// and is not changed here.
 			$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-			$product->set_manage_stock( false );
```
iamdharmesh, line 427: Shouldn't we set manage stock false for the genuine tracking off update?

includes/Handlers/Product.php
```diff
@@ -480,12 +519,27 @@ public static function update_products_stock_from_square( $product_ids ) {
...
+
+					if ( ! $changed ) {
+						continue;
+					}
+				} elseif ( ! $is_inventory_tracking ) {
+
+					// Not tracked in Square: reflect availability only; manage_stock is merchant
+					// intent and is not changed here.
 					$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-					$product->set_manage_stock( false );
```
iamdharmesh, line 488: same here.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1542,7 +1548,7 @@ protected function push_inventory() {
 
 						$inventory_change = Product::get_inventory_change_physical_count_type( $product );
 
-						if ( $inventory_change && $product->get_manage_stock() ) {
```
iamdharmesh, line 1545: As far as I know this is for the pushing count only and we only need to push the changes when the manage stock is on, should we push the inventory count when manage stock disabled as well?

iamdharmesh (commented):
Thanks for the great work here, @faisal-alvi! Overall, this looks good. I've added a few questions, and Copilot has also raised a few comments.

I haven't gone through a detailed review yet, but here are a couple of initial questions that I think are worth discussing:

1. The first test, **"Scenario A. The reported bug (stock wiped by a meaningless zero),"** worked well for me. When the merchant enables inventory sync afterward, the WooCommerce stock is no longer overwritten or reset to zero, which is great. However, after placing an order, the product quantity is reduced in WooCommerce and the Square inventory is updated to `-1`, creating a drift between the Square and WooCommerce inventory. Is there a way to update the Square inventory to match the WooCommerce inventory the first time inventory is pushed after inventory sync is enabled?

2. I've also added comments on this in the PR. The **Manage stock** option never gets disabled, even when inventory tracking is genuinely disabled on the Square side. If it was initially enabled in WooCommerce, it remains checked. Is this intentional?

Thanks

includes/Handlers/Product.php
faisal-alvi: Good question, this was a deliberate trade and here is the full picture for the genuine tracking off case.

**What happens when a merchant genuinely turns tracking off in Square:** the pull still reflects availability (stock status follows sold_out), but manage stock and the quantity stay whatever they are in WooCommerce. Nothing breaks: the product keeps selling and decrementing from its own WooCommerce settings, and there is no drift because an untracked Square item has no counts to disagree with. If the merchant's intent was "stop managing stock everywhere", they complete it by unticking Manage stock in WooCommerce, mirroring the conscious click they just made in Square. So the cost of removing the auto mirror is one manual click in that scenario.

**What happens when the quantity genuinely hits zero in Square (a real sellout, tracking stays on):** this path is fully preserved and is the core case the PR protects. Say a product with 10 in stock sells 10 times at the POS. Each sale writes an inventory adjustment; on the tenth, IN_STOCK reaches 0 and Square flags the item sold out. On the next pull (hourly poll, or immediately when a shopper triggers the add to cart refresh) the zero passes the history verification, because a real sellout always carries change records, and WooCommerce stock is written to 0. Manage stock stays on, and WooCommerce itself shows the product out of stock, since a managed product at quantity zero does that natively. Sold out on both sides, settings untouched. This is the "real zero written (77 to 0)" scenario in the PR test steps. The point of never touching manage stock is exactly that it is not needed for sellouts to land: quantity and status carry the state, the checkbox only carries the merchant's configuration.

**Why the auto mirror had to go:** the pull only receives a boolean, and `track_inventory = false` looks byte for byte identical whether the merchant toggled it on purpose or it is an artifact of an interrupted sync, a catalog wipe, or an item upserted while inventory sync was off. There is no provenance signal for tracking flags (change history covers counts, not flag flips), so the pull cannot tell intent from corruption. The artifact flavor is not hypothetical: on the SQUARE-364 customer export, a single poll on current trunk switched Manage stock off for 4,903 products through exactly this line, leaving them sellable with no stock control. One manual click in the rare intentional case against thousands of silent setting wipes in the failure case felt like the right side of the trade, and it is the same principle as the zero handling in this PR: a signal we cannot distinguish from corruption does not get write access to merchant settings.

**One refinement I am happy to add if you want it:** the argument above is strongest under WooCommerce SOR, where settings flow from WooCommerce to Square and a Square side toggle silently overriding them inverts the system of record. Under Square SOR the authority is reversed, so mirroring tracking off onto manage stock is defensible there. That would be a single `is_system_of_record_square()` condition, and the incident class we measured (WooCommerce SOR) stays fully protected. Let me know and I will include it in this round.

includes/Handlers/Product.php
faisal-alvi: Same answer as the other thread. Worth noting this second location is the automatic refresh that fires whenever a shopper adds the product to the cart, so with the old behavior a single customer browsing the store could wipe a product's stock settings if a failed sync had left the Square item untracked. That made this the most dangerous of the two call sites.

includes/Sync/Manual_Synchronization.php
faisal-alvi: The gate/condition did not disappear, I moved it inside `get_inventory_change_physical_count_type()`, which now decides from the product's full state:

- **Managed, quantity known:** push the real count (unchanged behavior).
- **Managed, quantity unresolved:** push nothing and log, where the old code fabricated a zero.
- **Not managed, out of stock:** push an explicit zero. This is the case that needs the gate removed here, and the reason is a Square API constraint: `sold_out` is read only and only becomes true through a tracked count of zero. Without this push there is no API level way to mark an unmanaged product sold out, so it stays sellable in Square forever. This was also the stuck sold out bug in reverse: a product that went out of stock once could never present as sold out, and one that came back in stock stayed sold out.
- **Not managed, in stock:** push nothing. The builder returns null, so keeping the old get_manage_stock() condition on this line would be redundant for this case and would block the out of stock case above.

So unmanaged products only ever push in one situation, the out of stock transition, and only an explicit zero. There is no path where we push a made up count for a product the merchant does not manage. The "unmanaged + outofstock pushes 0" and "unmanaged + instock pushes nothing" rows are both covered in the PR test steps.

Let me know your thoughts.

includes/Handlers/Product.php
faisal-alvi: Update after the later commits: For an item Square does not track, availability still follows `sold_out` exactly as before. For a tracked item reporting zero, a product that does not manage stock is now marked out of stock only when the zero is proven by inventory history; an unproven zero leaves it untouched, and a zero is never a reason to force it back in stock. That closes the mirror image of this ticket's bug, where a phantom zero was silently marking such products out of stock on every poll.

includes/Sync/Helper.php
```diff
@@ -89,6 +93,318 @@ public static function get_catalog_objects_tracking_stats( $catalog_object_ids )
...
+					$with_history[ $object_id ] = true;
+				}
+
+				$unresolved = array_diff( $chunk, array_keys( $with_history ) );
+
+				if ( empty( $unresolved ) || empty( $page['cursor'] ) ) {
+					continue;
+				}
+
+				foreach ( $unresolved as $object_id ) {
```
iamdharmesh: I think we should fetch inventory changes for the all $unresolved together instead looping through every item. what do you think?

includes/Sync/Helper.php
```diff
@@ -89,6 +93,318 @@ public static function get_catalog_objects_tracking_stats( $catalog_object_ids )
...
+					$single = self::fetch_inventory_changes_page( array( $object_id ) );
+					if ( null === $single ) {
+						return null;
```
iamdharmesh, line 234: Maybe continue; to allow next batch?

includes/Sync/Manual_Synchronization.php
iamdharmesh: > The gate/condition did not disappear, I moved it inside get_inventory_change_physical_count_type(), which now decides from the product's full state

Thanks for clarification.

includes/Sync/Manual_Synchronization.php
```diff
@@ -2151,7 +2196,10 @@ protected function assign_next_steps() {
 				// only handle product inventory if enabled
 				if ( wc_square()->get_settings_handler()->is_inventory_sync_enabled() ) {
 					$next_steps[] = 'push_inventory';
-					$next_steps[] = 'pull_inventory';
+					// pull_inventory is intentionally NOT queued when WooCommerce is the system of
+					// record: it only re-read the counts this same job just pushed, and Square's
+					// eventual consistency meant a not-yet-propagated count could come back as 0
+					// and overwrite the stock pushed moments earlier (SQUARE-145).
```
iamdharmesh: I think we should keep the "pull_inventory" step, but just shouldn't add the product IDs of the pushed products to "pull_inventory_variation_ids". So, we don't pull the pushed inventory again. The Manual sync not always pushes products and it also updates the existing products in that case. The  "pull_inventory" step is needed. However, we can get rid of this if we are pulling the inventory already as part of updating products. Am I missing anything here?

includes/Sync/Interval_Polling.php
```diff
@@ -310,30 +327,67 @@ protected function update_inventory_tracking() {
...
+				// Attempts exhausted: proceed with no zero verified, so only the zero writes are
+				// skipped and the job can finish.
+				$verified_zero_ids = array();
+
+				// Remember the objects so they are re-checked by id on later polls: the watermark can
+				// then advance normally without this window being lost.
+				Helper::remember_unverified_zero_counts( $zero_object_ids );
+			} else {
+				$this->clear_unverified_zero_count_attempts( 'update_inventory_tracking' );
+			}
```
iamdharmesh: I am seeing this code block at multiple places can we move this to helper function? or maybe as a part of get_catalog_objects_inventory_stats?

includes/Handlers/Product.php
iamdharmesh: Thanks for the explanation @faisal-alvi. It make sense to not sync the manage stock/tracking. However, it changes the behaviour from what we have currently. So, I think we need to pass this through the product approval once (cc: @builtbytay). 

> but manage stock and the quantity stay whatever they are in WooCommerce. Nothing breaks: the product keeps selling and decrementing from its own WooCommerce settings, and there is no drift because an untracked Square item has no counts to disagree with.

This may result into item become out of stock on WooCommerce and In stock on Square (As it is not tracking).

>  If the merchant's intent was "stop managing stock everywhere", they complete it by unticking Manage stock in WooCommerce, mirroring the conscious click they just made in Square. So the cost of removing the auto mirror is one manual click in that scenario.

Considering lots of products this manual click is not always feasible.


Also, we may want to flag this to Square team if they have a completed the webhook work and we can use it for get rid of all these workaround.

iamdharmesh (commented):
Thanks for the changes @faisal-alvi. I have added few question to discuss. I think it would be better to discuss this on call for the quick resolution whenever you are available. Thanks!

includes/Handlers/Product.php
builtbytay: @iamdharmesh @faisal-alvi let's keep the current behavior — don't auto-untick Manage stock. An unrecoverable mass wipe (SQUARE-364, 4,903 products) is a worse outcome than merchants unticking manually, and we can't tell a real toggle from a sync failure right now.

We should take the merchant-effort problem as a separate ticket: mirroring tracking-off safely needs either a Square SOR condition or a webhook that confirms the change is intentional. 

When we release this, let's also make sure to let the Woo support team know about the manual step in Slack so that we can potentially reduce the amount of tickets that come in related to that decision. cc @christinahu-cmd

includes/Sync/Manual_Synchronization.php
faisal-alvi: `pull_inventory` only reads `pull_inventory_variation_ids`, and in Woo SOR mode exactly one thing writes it: the loop over the upsert response in `upsert_catalog_objects()`. The other writer is `square_sor_sync()`, which only runs when Square is the system of record, and there the step is untouched.

On the "it also updates existing products" case: `update_matched_products()` and `search_matched_products()` both call `upsert_catalog_objects()`, which is the method that issues `batch_upsert_catalog_objects`. So those products are pushed products too, and in Woo SOR there is no id in that list this same job did not just push. That makes your suggestion and this change behave identically, since with the pushed ids excluded the list is empty and the step returns at its own empty guard, except it would still cost a step cycle per sync.

Your comment did surface something real though. `inventory_push_product_ids` is only ever written inside `upsert_new_products()`, which subtracts `processed_product_ids`, and both matched steps fill that. So on a repeat sync `push_inventory` runs with an empty list and pushes nothing, meaning a Woo SOR full sync never re-pushes quantities for products that already exist in Square. That is a genuine gap and I would rather fix it deliberately in its own ticket, since for a large catalog it means thousands of physical count writes per full sync and needs its own rate limit handling and testing.

includes/Sync/Helper.php
faisal-alvi: Can't, because an id missing from the result already means "checked, no history". Skip a failed one and it looks verified, so its zero is dropped as a phantom and nothing retries it. Returning null makes the step retry the batch, which is what we want here.

faisal-alvi:
@iamdharmesh thanks for the review, and for the call. As per the feedback on a call, the retry logic around unverified zero counts felt over engineered, and decided we either drop it or simplify it and just notify the merchant about those counts. That is done, please have a look and let me know what you think.

includes/Sync/Interval_Polling.php
```diff
@@ -37,6 +37,10 @@
  */
 class Interval_Polling extends Stepped_Job {
 
+
+
+
+
```
iamdharmesh: Nit: Remove this empty space.

includes/Sync/Helper.php
iamdharmesh: Thanks for the clarification.

includes/Handlers/Product/Woo_SOR.php
```diff
@@ -412,26 +412,52 @@ public static function update_catalog_variation( CatalogObject $catalog_object,
...
 
 			/*
-			 * Only update track_inventory if it's not set.
+			 * Only update the base track_inventory if it's not set.
 			 * This will only update inventory tracking on new variations.
 			 * inventory tracking will remain the same for existing variations.
 			 */
 			if ( is_null( $track_inventory ) && is_null( $location_overrides ) ) {
 				$variation_data->setTrackInventory( $product->get_manage_stock() );
+			}
```
iamdharmesh, line 421: With this early return we are now pushing the track inventory every time instead of the first time when product is pushed. I think this should be pushed first time only. am I missing anything here?

includes/Sync/Manual_Synchronization.php
iamdharmesh: > On the "it also updates existing products" case: `update_matched_products()` and `search_matched_products()` both call `upsert_catalog_objects()`, which is the method that issues `batch_upsert_catalog_objects`.

@faisal-alvi for the update_matched_products()` and `search_matched_products()` both product get pushed but the inventory get pulled for the updated products due to INVENTORY ALWAYS PULLED FROM SQUARE rule and removing this will completely removing `pull_inventory` step with remove the inventory sync for the Woo SOR manual sync. So, this step is still mandatory.

includes/Handlers/Product.php
iamdharmesh: Hi @faisal-alvi, I just noticed when "Sync with Square" checked on product we are disabling the checkbox/toggle for the manage stock on the product. So, merchant won't be able to this manually, We have to provide a way to merchant to toggle this and docs update regarding this as well to clarify this to support team and merchants as well.

iamdharmesh (commented):
Thanks for the updates around verifying the zero counts. I haven't fully tested this yet, but the code changes look good to me. I have added some more comments mostly on the previous discussion. Could you please help to check on that?

faisal-alvi:
@iamdharmesh Thanks for the feedback. Back to you for review. I added a few commits; let's discuss on Slack/Call if you need any clarification.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1880,6 +1891,45 @@ function ( $square_product_variation ) {
...
+		}
+
+		if ( ! $object_ids ) {
+			return;
+		}
+
+		$this->set_attr(
+			'pushed_inventory_variation_ids',
+			array_values( array_unique( array_merge( (array) $this->get_attr( 'pushed_inventory_variation_ids', array() ), $object_ids ) ) )
+		);
```
iamdharmesh: Nit: Instead of setting `pushed_inventory_variation_ids` and skip it in pull_inventory step, can't we directly update the "pull_inventory_variation_ids" to remove the pushed variation_ids? what do you think?

includes/Handlers/Product.php
iamdharmesh: Thanks for the enabling the checkbox/toggle for the manage stock on the product @faisal-alvi. With this enabled now I can see that merchant uncheck the manage stock on the product also get updated to the Square side as well. However, checking it back not reflect it back on the Square side (which I believe correct). But could you please help me list out the Woo side update scenario with expected behaviour to make sure we all are on same page and we can get feedback from Taylor if needed. Thanks.

> One refinement I am happy to add if you want it: the argument above is strongest under WooCommerce SOR, where settings flow from WooCommerce to Square and a Square side toggle silently overriding them inverts the system of record. Under Square SOR the authority is reversed, so mirroring tracking off onto manage stock is defensible there. That would be a single is_system_of_record_square() condition, and the incident class we measured (WooCommerce SOR) stays fully protected. Let me know and I will include it in this round.

I'd love to add this. seems you missed but could you please add it?

iamdharmesh (commented):
Thanks for the changes @faisal-alvi. Adding few last comments to make sure we have addressed all raised feedback. Could you please help to check once?

> The first test, "Scenario A. The reported bug (stock wiped by a meaningless zero)," worked well for me. When the merchant enables inventory sync afterward, the WooCommerce stock is no longer overwritten or reset to zero, which is great. However, after placing an order, the product quantity is reduced in WooCommerce and the Square inventory is updated to -1, creating a drift between the Square and WooCommerce inventory. Is there a way to update the Square inventory to match the WooCommerce inventory the first time inventory is pushed after inventory sync is enabled?

It seems we missed this once. Could you please help to check once?

includes/Sync/Manual_Synchronization.php
faisal-alvi: Good call, done

includes/Handlers/Product.php
faisal-alvi: Added. Under Square SOR, tracking off in Square now also switches Manage stock off on the product. Under WooCommerce SOR nothing changes, so a Square side toggle cannot silently override the merchant's setting.</p><p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext;">Woo side scenarios, as they now behave:</p>
Merchant does in Woo | Under Woo SOR | Under Square SOR
-- | -- | --
Unticks Manage stock | Stays off. Pushed to Square as tracking off | Stays off, and Square is authoritative anyway
Ticks Manage stock back on | Stays on in Woo, not pushed to Square | Reverted to off on the next pull if Square still has tracking off
Unticks with Square still tracking | Reverted on the next pull, since a positive count mirrors | Same
Unticks "Sync with Square" | Square stops touching the product entirely, on all four pull paths | Same

<p style="white-space: pre-wrap; margin-top: 0.1em; margin-bottom: 0.2em; unicode-bidi: plaintext;">The asymmetry you spotted, unticking reaching Square but re-ticking not, is because the base tracking flag is only written when Square has none, so we never turn tracking back on for an existing item. Happy to change that if @builtbytay wants symmetry, but it is a deliberate "do not re-enable what the merchant turned off in Square" rule.</p>

faisal-alvi:
@iamdharmesh thanks, all three addressed.

**Pull queue nit:** done your way. Pushed objects are removed from `pull_inventory_variation_ids` at push time, so the extra attribute and the skip logic in `pull_inventory()` are gone. It relies on both push sites running before `pull_inventory` in the step order, which I have noted in the code.

**Square SOR refinement:** added. When Square is the system of record, tracking off in Square now also switches Manage stock off on the product. Under WooCommerce SOR nothing changes, so a Square side toggle cannot override the merchant's setting. Full scenario table is in the thread on Product.php.

**The drift you found:** real gap, fixed at the cause. An order sends a relative adjustment, which assumes Square already holds a count. With no count, the -1 delta became Square's entire count, hence Woo at 19 and Square at -1. Items Square has never counted now receive WooCommerce's current quantity as an absolute count on the first order, establishing the baseline, and later orders adjust from it normally. If the history check is unavailable we keep the previous behaviour rather than guessing.

includes/Handlers/Product.php
```diff
@@ -480,12 +532,27 @@ public static function update_products_stock_from_square( $product_ids ) {
...
```
iamdharmesh, line 488: Should update the manage stock when Square is SOR?
```php
if ( wc_square()->get_settings_handler()->is_system_of_record_square() ) {
	$product->set_manage_stock( false );
}
```

includes/Sync/Interval_Polling.php
```diff
@@ -391,45 +425,66 @@ protected function update_inventory_counts() {
...
+					);
+
+					if ( ! $changed ) {
+						continue;
+					}
 				} else {
+					// Not tracked in Square: reflect availability only; manage_stock is merchant
+					// intent and is not changed here.
 					$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-					$product->set_manage_stock( false );
```
iamdharmesh, line 432: Same here.

includes/Sync/Manual_Synchronization.php
```diff
@@ -2004,23 +2073,33 @@ protected function pull_inventory() {
...
-				} else {
+					if ( $changed ) {
+						$product->save();
+					}
+				} elseif ( ! $is_tracking_inventory ) {
+
+					// Not tracked in Square: reflect availability only; manage_stock is merchant
+					// intent and is not changed here.
 					$product->set_stock_status( $sold_out ? 'outofstock' : 'instock' );
-					$product->set_manage_stock( false );
```
iamdharmesh, line 2020: Same here.

src/js/admin/wc-square-admin-products.js
```diff
@@ -415,10 +415,9 @@ jQuery( document ).ready( ( $ ) => {
 							syncInventory
 					);
 
-					$manageInput.on('click', () => {
-						return false;
-					});
-					$manageInput.css({ opacity: '0.5' });
```
iamdharmesh, line 421: Nit:
As we are updating "manage stock" for Square SOR, we should keep this disabled for the Square SOR.

iamdharmesh (approved):
Thanks for the update here @faisal-alvi. This looks good to me now. I’m approving with the following notes:

- We should document the behavior change to ensure that merchants and the support team are aware of it (thread: https://github.com/woocommerce/woocommerce-square/pull/529#discussion_r3734997486)
- For the future, we should consider adding caching or storing metadata on the product to avoid requesting inventory history every time.

@builtbytay, this needs one check from you as well: https://github.com/woocommerce/woocommerce-square/pull/529#discussion_r3850494097

qasumitbagthariya:
### QA Update  ✅
---

I have verified this PR in the `fix/SQUARE-145-zeroing` branch, which has been fixed and is functioning as intended.

Verified all inventory sync scenarios, including zero-stock handling, stock management, Square sellouts, variations, sync exclusions, push/pull inventory, order reconciliation, and Square-as-system-of-record imports.

All scenarios passed successfully with no regressions observed. ✅

🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/SQUARE-145-zeroing

</Details>


🔁 Next Step  -  Ready for UAT


Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

qasumitbagthariya (approved):

PR #530, opened 2026-08-05
Prevent duplicate Square card field when a checkout update lands during form initialization

petenelson:
### All Submissions:

> The new E2E spec is written but has **not** been run locally — it needs the wp-env environment plus Square sandbox credentials. Relying on CI for that. The fix itself was verified headlessly against the compiled bundle (see below).

---

Fixes a race condition in `set_payment_fields()` that renders the Square credit card field **twice** on classic (shortcode) checkout, blocking checkout for affected shoppers.

**Root cause.** `set_payment_fields()` guards all of its teardown/reuse logic on `this.payment_form`, but that property is only assigned *inside* the async `payments.card()` promise, after `attach()`. Between the first call and that promise resolving, `this.payment_form` is `undefined`. A second `updated_checkout` landing in that window falls straight past the guard, skips the destroy, and calls `initializeCard()` again — attaching a second card iframe into the same `#wc-square-credit-card-container`.

Any plugin that fires one `update_checkout` on page load triggers it (confirmed on the reporting merchant's site: CodeRockz "Woo Delivery"; also reported publicly against Order Delivery Date for WooCommerce Lite). Despite the issue title, this is **not** Chrome-specific — it is page-load timing, and "Chrome but not incognito/Edge/Firefox" are proxies for load speed.

Three related holes in the same lifecycle are closed too:

1. **`attach()` was not awaited.** `this.payment_form` was published before the iframe mounted, leaving a second window where the "already initialized" guard is false while `payment_form` is truthy — which pushes a concurrent update into the *destroy* branch.
2. **The destroy branch had the same re-entrancy hole.** `this.payment_form` was never cleared while `destroy()` was in flight, so a further update would destroy an already-destroyed form and re-initialize — another double-attach path.
3. **`payments.card()` had no rejection handler.** An SDK failure produced an unhandled promise rejection, left the perf metric open, and gave the shopper no card field and no recovery path. This is a plausible cause of the "payment nonce is missing / invalid card on first attempt, retry succeeds" reports in Zendesk #11475839.

**The fix** (`src/js/frontend/wc-square.js`):

- A `settingUp` re-entrancy flag, set **synchronously** before `payments.card()` and cleared in a `finally`. A checkout update arriving mid-build sets `pending_refresh` and returns rather than starting a competing build. Mirrors the existing `this.settingUp` pattern in `wc-square-cash-app-pay.js`.
- `pending_refresh` is flushed once the build settles, so a postcode or total change that landed during the window is still applied (it re-enters and hits the cheap `configure()` branch).
- `await card.attach()` before assigning `this.payment_form`, so the handler is never exposed pointing at an unmounted card.
- New `rebuild_payment_form()` nulls `this.payment_form` before awaiting `destroy()`.
- The container is emptied before attaching — a defensive second layer that also covers any other double-attach path. Safe because `includes/Gateway/Payment_Form.php:147` renders the div empty; it is a pure mount point.
- The whole build is wrapped in `try/catch/finally`, so a failed build logs and recovers instead of latching the flag.

Blocks checkout is unaffected (separate React path in `src/blocks/credit-card/`).

Closes SQUARE-355 — https://linear.app/a8c/issue/SQUARE-355/duplicate-credit-card-field-rendered-on-checkout-in-chrome-classic

Deterministic reproduction, from the investigation on the Linear issue — drop this into Code Snippets (PHP, "Run everywhere") on a store with classic shortcode checkout and Square enabled. It is opt-in by query param, so checkout behaves normally without one:

```php
add_action( 'wp_footer', function () {
	if ( ! function_exists( 'is_checkout' ) || ! is_checkout() ) {
		return;
	}
	if ( ! isset( $_GET['ted_mode'] ) ) {
		return;
	}
	?>
	<script>
	jQuery( function ( $ ) {
		$( document.body ).one( 'updated_checkout', function () {
			$( document.body ).trigger( 'update_checkout' );
		} );
	} );
	</script>
	<?php
}, 100 );
```

1. Add a product to the cart and visit the **classic shortcode** checkout with `?ted_mode=onupdate`.
2. Select Square Credit Card. In the console, run `document.querySelectorAll('#wc-square-credit-card-container iframe.sq-card-component').length`.
   * On `trunk` this returns **2** and two card fields are visible.
   * On this branch it returns **1**.
3. Reload without the query param (control) — still exactly one field, and it renders normally.
4. Change the shipping method or country a few times to force several `updated_checkout` cycles — still exactly one field each time.
5. Complete the order with test card `4111 1111 1111 1111`. It should tokenize and place the order with no errors.
6. Repeat in Firefox, Safari and Edge to confirm nothing regressed.

Alternatively, the staging site from the Linear investigation reproduces on `trunk` without any setup:
`https://squarefeten.mystagingwebsite.com/checkout/?ted_mode=onupdate`

### Verification already done

The compiled bundle was driven headlessly in jsdom against a stubbed Square SDK, modelling classic checkout's payment-box replacement. Both bundles built from the same webpack config:

| Scenario | `trunk` | This branch |
| --- | --- | --- |
| Control — single checkout update | 1 iframe | 1 iframe |
| Race — refresh during card build | **2 iframes, 2 wrappers** | 1 iframe |
| Burst — 4 refreshes, 25 ms apart | **3 iframes** | 1 iframe |
| Sequential — 3 settled refreshes | 1 iframe | 1 iframe |
| SDK rejects the first build | **unhandled rejection, no recovery** | recovers, 1 iframe |

A sweep across five `payments.card()` delays duplicated **4/5 on `trunk` and 0/5 on this branch**, and in the racing cases only *one* card object is ever created — the redundant build is prevented, not merely cleaned up afterwards. In every passing case `payment_form` ends up set and attached, which is what `handleSubmission()` needs in order to tokenize.

Two caveats on that evidence:

* The burst case duplicates in the harness but the browser-based investigation on the Linear issue found bursts *suppress* the bug in a real browser. Not a contradiction — the harness does not model WooCommerce aborting in-flight `update_order_review` XHRs, which is what collapses a real burst into a single late update. It is a harsher variant, not a refutation.
* The harness is a scratch tool and is not committed.

### Follow-ups (not in this PR)

* **SQUARE-301** is the same *class* of bug (duplication across `updated_checkout` cycles) on the gift card path.
* `src/blocks/credit-card/square-web-payments-form.js` has a narrower version of the same pattern — `card` state is set after the `await`, and `postalCode` is an effect dependency. React-guarded and not what was reported, but worth its own ticket.
* The issue title and description should drop "in Chrome"; it misdirects.

### Changelog entry

> Fix - Prevent a second Square credit card field from being rendered when a checkout update is triggered while the payment form is still initializing.

### DIFF
--- a/src/js/frontend/wc-square.js
+++ b/src/js/frontend/wc-square.js
@@ -51,6 +51,12 @@ jQuery( document ).ready( ( $ ) => {
 				args.is_change_payment_method_request;
 			this.metrics = {};
 
+			// Whether a payment form build is currently in flight.
+			this.settingUp = false;
+
+			// Whether a checkout update landed while a build was in flight.
+			this.pending_refresh = false;
+
 			if ( $( 'form.checkout' ).length ) {
 				this.form = $( 'form.checkout' );
 				this.handle_checkout_page();
@@ -200,6 +206,14 @@ jQuery( document ).ready( ( $ ) => {
 		 * @since 2.0.0
 		 */
 		set_payment_fields() {
+			// A build is already in flight. `this.payment_form` isn't assigned until it settles, so
+			// carrying on here would skip the guards below and attach a second card to the container.
+			// Queue a single refresh instead - it's flushed once the build completes.
+			if ( this.settingUp ) {
+				this.pending_refresh = true;
+				return;
+			}
+
 			if ( this.payment_form ) {
 				// Don't re-initialize the payment form if it's already been initialized and exist in DOM.
 				if (
@@ -215,10 +229,13 @@ jQuery( document ).ready( ( $ ) => {
 
 				// Destroy the payment form and re-initialize it.
 				this.log( 'Destroying payment form' );
-				this.payment_form.destroy().then( () => {
-					this.log( 'Re-building payment form' );
-					this.initializeCard( this.payments );
-				} );
+
+				const previous_payment_form = this.payment_form;
+
+				// Drop the reference before awaiting, so a checkout update landing mid-destroy
+				// can't tear the same form down twice.
+				this.payment_form = null;
+				this.rebuild_payment_form( previous_payment_form );
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


petenelson:
### CI status: the red `E2E tests (@sync)` check is pre-existing, not from this PR

Current state on `b2c3f95` — 8 of 9 checks pass:

| Check | Result |
| --- | --- |
| eslint / phpcs / PHP compatibility / test (Plugin Check) | pass |
| E2E `@general` | **pass** — 59 passed, 0 failed |
| E2E `@cashapp` | pass |
| E2E `@giftcard` | pass |
| E2E `@sync` | **fail** |

#### Why `@sync` is unrelated to this change

This branch has had two E2E runs. The only difference between them is the new E2E spec file — **no plugin code changed** between the two. Yet the set of failing `@sync` tests changed completely:

| Run | Commit | Failing `@sync` tests |
| --- | --- | --- |
| [31027004449](https://github.com/woocommerce/woocommerce-square/actions/runs/31027004449) | `8b8c56f` | `c2.woo-sor`, `c3.woo-sor-with-inventory`, `c4.product-import` |
| [31038627935](https://github.com/woocommerce/woocommerce-square/actions/runs/31038627935) | `b2c3f95` | `d9.order-sync-fulfillment` |

Different failures from identical checkout code across two runs is non-determinism, not a regression. All of these are admin-side Square catalog and order sync specs — no checkout JS is involved in any of them, and this PR touches only `src/js/frontend/wc-square.js`.

`@sync` also fails on other recent branches, e.g. [fix/SQUARE-167](https://github.com/woocommerce/woocommerce-square/actions/runs/30825443017) and [fix/SQUARE-291](https://github.com/woocommerce/woocommerce-square/actions/runs/30550910751).

#### `@giftcard` flaked, then went green

`@giftcard` failed on the first run with a single test — `b8.gift-card-product.spec.js:128 › Gift card recipient email`, which asserts on rendered **email template** text. The identical test fails all three retries on [fix/SQUARE-145-zeroing](https://github.com/woocommerce/woocommerce-square/actions/runs/30930217787), an unrelated branch. It passed on the second run here with no gift card code changed in between.

Worth noting for reviewer confidence: every gift card *payment* test passed in both runs — full payment, partial payment, partial payment + refund, and gift card product purchase — and all of those drive the credit card form this PR modifies.

#### What the passing `@general` run actually proves

Both new tests pass:

```
✓ 22 a94.duplicate-credit-card-field.spec.js:107 › Credit card field is not duplicated by a checkout update during form init (5.2s)
✓ 23 a94.duplicate-credit-card-field.spec.js:120 › Order can be placed after a checkout update during form init (12.7s)
```

The second commit on this branch exists specifically to make that green meaningful. The first version of the spec could have passed without ever exercising the bug, and a retrying `toHaveCount(1)` matcher would pass the instant the count touched 1 — which on a duplicating build is briefly true before the second card attaches. The spec now waits on an in-page `updated_checkout` counter (`waitForFunction(count >= 2)`), so it fails outright if the second checkout cycle never fires, and then takes a single non-retrying snapshot of the iframe and wrapper counts.

So the pass confirms the race genuinely fired and the card count still settled at exactly 1.

#### Unrelated, but worth a ticket

7 of the last 8 non-cancelled E2E runs are red across six unrelated branches. The suite looks chronically flaky independent of any of these PRs.

PR #538, opened 2026-08-14
Prevent product sync from getting stuck when a product set to sync with Square is deleted.

### LINKED ISSUE #539: Prevent product sync from getting stuck when a product set to sync with Square is deleted.
PR https://github.com/woocommerce/woocommerce-square/pull/538 did not have a linked issue, so this issue has been created for tracking purposes.

iamdharmesh:
### All Submissions:

---

Product sync could fatal and get stuck when a product set to sync with Square no longer exists. When the sync job processed an ID that `wc_get_product()` could not resolve, the code called methods on the result (`get_type()`, `get_meta()`), causing a fatal error that halted the job and blocked the sync queue.

This handles the missing-product case at both points where it can occur:

- **`validate_products()`** — covers a product that is already gone when the sync starts. The ID is skipped (with a merchant-facing alert) instead of reading its type.
- **`upsert_catalog_objects()`** — covers a product that still existed at validation time but is permanently deleted **while a long sync is running** (the job is stepped/batched across many background runs, so there is a real window between validating an ID and upserting it). Because `upsert_new_products()` builds its batch from the validated IDs without re-fetching, this is where a mid-sync deletion surfaces.

Rather than only skipping the missing product in `upsert_catalog_objects()` (which would leave it in the "unprocessed" set forever and keep the sync step from ever completing), the missing IDs are tracked and excluded from the unprocessed set so the step can finish. The tracking is stored with the in-progress upsert data so a resumed run still excludes them, and an empty batch (every product in it deleted) is short-circuited instead of sending an empty request to Square. Each skipped product is logged once for troubleshooting.

**Backward compatibility:** No BC impact — the changes are internal guards and bookkeeping inside two `protected` methods; no public signature, hook, template, option, or REST contract is affected.

**A. Product already deleted before sync (validate path)**
1. Connect the store to Square with product sync enabled (WooCommerce as the system of record).
1. Create/sync a product so it is set to sync with Square, then permanently delete it (Trash → Delete permanently).
1. Confirm the sync completes without a fatal error and the queue is not stuck, and an alert appears on the sync screen noting the product was excluded because it no longer exists. (Currently sync get stuck on `trunk` branch).

**B. Product deleted while a sync is running (upsert path)**
1. With product sync enabled, start a sync by clicking a "Sync now"(a larger batch makes the window easier to hit).
1. While the sync is in progress, permanently delete one of the products being synced.
1. Let the sync run to completion.
1. Confirm the sync finishes without a fatal error and the remaining products sync successfully — rather than the sync stalling or re-running the same step indefinitely. The skipped product is noted once in the Square debug log (WooCommerce → Status → Logs, `wc-square` handle).

### Changelog entry

> Fix - Prevent product sync from getting stuck when a product set to sync with Square is deleted.

Closes #539

### DIFF
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -1034,13 +1034,6 @@ class Manual_Synchronization extends Stepped_Job {
 
 		$result = $this->upsert_catalog_objects( $catalog_objects, true );
 
-		// Track products that no longer resolve to a WC_Product (deleted mid-sync). They are already
-		// recorded with a dedicated "excluded from sync" alert inside upsert_catalog_objects(), so
-		// accumulate them across batches and keep them out of the "could not be updated" report below
-		// - a deletion should not be surfaced as an update failure.
-		$missing_product_ids = array_values( array_unique( array_merge( $this->get_attr( 'upsert_missing_product_ids', array() ), $result['missing'] ?? array() ) ) );
-		$this->set_attr( 'upsert_missing_product_ids', $missing_product_ids );
-
 		// Push inventory inline immediately after each upsert batch when inventory sync is enabled.
 		// This ensures products don't sit in Square with zero inventory if the sync fails before
 		// the deferred push_inventory step runs. Only IDs whose inline push failed are queued for
@@ -1067,6 +1060,7 @@ class Manual_Synchronization extends Stepped_Job {
 			// at this point, log a failure for any products that weren't processed. Products deleted
 			// mid-sync are excluded - they already have their own "excluded from sync" alert and were
 			// never processable, so reporting them as update failures would be misleading.
+			$missing_product_ids = $this->get_attr( 'upsert_missing_product_ids', array() );
 			foreach ( array_diff( $all_product_ids, $processed_product_ids, $missing_product_ids ) as $product_id ) {
 				Records::set_record(
 					array(
@@ -1115,7 +1109,6 @@ class Manual_Synchronization extends Stepped_Job {
 			'in_progress_upsert_catalog_objects',
 			array(
 				'staged_product_ids'                => array(),
-				'missing_product_ids'               => array(),
 				'unprocessed_upsert_response'       => null,
 				'mapped_client_item_ids'            => array(),
 				'processed_remote_catalog_item_ids' => array(),
@@ -1123,10 +1116,11 @@ class Manual_Synchronization extends Stepped_Job {
 		);
 
 		// Product IDs that no longer resolve to a WC_Product (e.g. permanently deleted mid-sync).
-		// Tracked so they can be dropped from the unprocessed set below, otherwise they would be
-		// re-queued forever and the sync step would never complete. Restored from in-progress data
-		// so a resumed run (which skips the staging loop) still excludes them.
-		$missing_product_ids = $in_progress['missing_product_ids'] ?? array();
+		// Tracked in a job-level attribute so they can be dropped from the unprocessed set below,
+		// otherwise they would be re-queued forever and the sync step would never complete. The set
+		// is cumulative across batches, sync steps (update_matched_products, search_matched_products,
+		// upsert_new_products), retries, and resumed runs, so each product is handled exactly once.
+		$missing_product_ids = $this->get_attr( 'upsert_missing_product_ids', array() );
 
 		$upsert_response = null;
 		if ( ! empty( $in_progress['unprocessed_upsert_response'] ) ) {
@@ -1148,11 +1142,14 @@ class Manual_Synchronization extends Stepped_Job {
 				$product = wc_get_product( $product_id );
 
 				if ( ! $product instanceof \WC_Product ) {
-					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
-					// out of the unprocessed set and the step can complete rather than re-queueing it
-					// forever. Logged and recorded once per product (and not repeated on a resumed run).
+					// Product is gone (e.g. permanently deleted mid-sync). Record it in the job-level
+					// missing set so it drops out of the unprocessed set and the step can complete rather
+					// than re-queueing it forever. The set is persisted and cumulative, so the guard also
+					// ensures the log entry and merchant-facing alert fire exactly once per product per
+					// job even if it is encountered again in a later step, a retried batch, or a resume.
 					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
 						$missing_product_ids[] = $product_id;
+						$this->set_attr( 'upsert_missing_product_ids', $missing_product_ids );
 
 						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );
 
@@ -1237,7 +1234,6 @@ class Manual_Synchronization extends Stepped_Job {
 			}
 
 			$in_progress['staged_product_ids']          = $staged_product_ids;
-			$in_progress['missing_product_ids']         = $missing_product_ids;
 			$in_progress['unprocessed_upsert_response'] = wp_json_encode( $upsert_response, JSON_PRETTY_PRINT );
 			$this->set_attr( 'in_progress_upsert_catalog_objects', $in_progress );
 
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -109,6 +109,16 @@ class Manual_Synchronization extends Stepped_Job {
 			foreach ( $unsupported_product_ids as $matched_product_id ) {
 				$product = wc_get_product( $matched_product_id );
 				if ( ! $product instanceof \WC_Product ) {
+					Records::set_record(
+						array(
+							'type'    => 'alert',
+							'message' => sprintf(
+								/* translators: %1$s - Product ID. */
+								__( 'Product ID (%1$s) is excluded from sync as the product not found.', 'woocommerce-square' ),
+								$matched_product_id,
+							),
+						)
+					);
 					continue;
 				}
 
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -1102,7 +1102,6 @@ class Manual_Synchronization extends Stepped_Job {
 		$result                    = array(
 			'processed'   => array(),
 			'unprocessed' => $product_ids,
-			'missing'     => array(),
 		);
 
 		$in_progress = $this->get_attr(
@@ -1192,7 +1191,6 @@ class Manual_Synchronization extends Stepped_Job {
 				$this->set_attr( 'in_progress_upsert_catalog_objects', null );
 
 				$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
-				$result['missing']     = $missing_product_ids;
 
 				return $result;
 			}
@@ -1375,10 +1373,9 @@ class Manual_Synchronization extends Stepped_Job {
 
 		$result['processed'] = $staged_product_ids;
 		// Exclude missing products (deleted mid-sync) from unprocessed so callers do not re-queue
-		// them indefinitely, which would keep the sync step from ever completing. They are also
-		// returned so callers can tell "deleted" apart from "failed to update".
+		// them indefinitely, which would keep the sync step from ever completing. The IDs themselves
+		// are tracked in the upsert_missing_product_ids job attribute for callers that need them.
 		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
-		$result['missing']     = $missing_product_ids;
 
 		return $result;
 	}
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -1105,11 +1105,18 @@ class Manual_Synchronization extends Stepped_Job {
 			'in_progress_upsert_catalog_objects',
 			array(
 				'staged_product_ids'                => array(),
+				'missing_product_ids'               => array(),
 				'unprocessed_upsert_response'       => null,
 				'mapped_client_item_ids'            => array(),
 				'processed_remote_catalog_item_ids' => array(),
 			)
 		);
+
+		// Product IDs that no longer resolve to a WC_Product (e.g. permanently deleted mid-sync).
+		// Tracked so they can be dropped from the unprocessed set below, otherwise they would be
+		// re-queued forever and the sync step would never complete. Restored from in-progress data
+		// so a resumed run (which skips the staging loop) still excludes them.
+		$missing_product_ids = $in_progress['missing_product_ids'] ?? array();
 
 		$upsert_response = null;
 		if ( ! empty( $in_progress['unprocessed_upsert_response'] ) ) {
@@ -1131,6 +1138,14 @@ class Manual_Synchronization extends Stepped_Job {
 				$product = wc_get_product( $product_id );
 
 				if ( ! $product instanceof \WC_Product ) {
+					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
+					// out of the unprocessed set and the step can complete rather than re-queueing it
+					// forever. Logged once per product (and not re-logged on a resumed run).
+					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
+						$missing_product_ids[] = $product_id;
+
+						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );
+					}
 					continue;
 				}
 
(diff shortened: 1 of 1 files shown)


includes/Sync/Manual_Synchronization.php
```diff
@@ -1114,7 +1135,20 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
 				}
 
-				$product                                  = wc_get_product( $product_id );
+				$product = wc_get_product( $product_id );
+
+				if ( ! $product instanceof \WC_Product ) {
+					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
+					// out of the unprocessed set and the step can complete rather than re-queueing it
+					// forever. Logged once per product (and not re-logged on a resumed run).
+					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
```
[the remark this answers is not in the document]
iamdharmesh: This is already fixed under https://github.com/woocommerce/woocommerce-square/pull/524

includes/Sync/Manual_Synchronization.php
```diff
@@ -108,7 +108,21 @@ protected function validate_products() {
...
 			foreach ( $unsupported_product_ids as $matched_product_id ) {
 				$product = wc_get_product( $matched_product_id );
-				$type    = $product->get_type();
+				if ( ! $product instanceof \WC_Product ) {
+					Records::set_record(
+						array(
+							'type'    => 'alert',
+							'message' => sprintf(
+								/* translators: %1$s - Product ID. */
+								__( 'Product ID (%1$s) is excluded from sync as the product not found.', 'woocommerce-square' ),
```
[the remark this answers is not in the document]
iamdharmesh: Fixed in https://github.com/woocommerce/woocommerce-square/pull/538/commits/0294b77dcabda751ec9852092d85bca30406c078

includes/Sync/Manual_Synchronization.php
```diff
@@ -1167,6 +1213,7 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
 			}
 
 			$in_progress['staged_product_ids']          = $staged_product_ids;
+			$in_progress['missing_product_ids']         = $missing_product_ids;
```
faisal-alvi: `missing_product_ids` is only persisted after the API call succeeds. If the batch throws (rate limit, transport), the IDs collected in that pass are lost and get re-detected and re-logged on the next run, so "logged once per product" holds only for the non-throwing path. Not worth extra state if you are fine with the duplicate log line - flagging so the description matches the behaviour.

includes/Sync/Manual_Synchronization.php
```diff
@@ -108,7 +108,21 @@ protected function validate_products() {
...
 				$product = wc_get_product( $matched_product_id );
-				$type    = $product->get_type();
+				if ( ! $product instanceof \WC_Product ) {
+					Records::set_record(
+						array(
+							'type'    => 'alert',
+							'message' => sprintf(
+								/* translators: %1$s - Product ID. */
+								__( 'Product ID (%1$s) is excluded from sync because the product could not be found.', 'woocommerce-square' ),
+								$matched_product_id,
```
faisal-alvi: Trailing comma after the last `sprintf()` argument, which the sibling `sprintf()` right below does not have. Fine on PHP 7.4+, purely consistency.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1306,8 +1353,10 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
 
 		$this->set_attr( 'in_progress_upsert_catalog_objects', null );
 
-		$result['processed']   = $staged_product_ids;
-		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids );
+		$result['processed'] = $staged_product_ids;
+		// Exclude missing products (deleted mid-sync) from unprocessed so callers do not re-queue
+		// them indefinitely, which would keep the sync step from ever completing.
+		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
```
faisal-alvi: Missing IDs are excluded from `unprocessed` here but never added to `processed_product_ids`, so the block at the end of `upsert_new_products()` that logs "any products that weren't processed" picks them up and records `Product #<a href="">11068</a> could not be updated.` - empty href, because `get_edit_post_link()` returns null once the post is gone, and wording that does not mention the deletion.

Returning the missing IDs lets the caller tell the two cases apart (caller-side diff is in the review summary, since those lines are not in this diff):

```suggestion
		$result['processed'] = $staged_product_ids;
		// Exclude missing products (deleted mid-sync) from unprocessed so callers do not re-queue
		// them indefinitely, which would keep the sync step from ever completing. They are also
		// returned so callers can tell "deleted" apart from "failed to update".
		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
		$result['missing']     = $missing_product_ids;
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -1114,7 +1135,20 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
+					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
+					// out of the unprocessed set and the step can complete rather than re-queueing it
+					// forever. Logged once per product (and not re-logged on a resumed run).
+					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
+						$missing_product_ids[] = $product_id;
+
+						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );
+					}
+					continue;
+				}
```
faisal-alvi: Two notes on this block.

First, confirmed the resumed-run case matters: a resumed run skips this whole staging loop, so restoring `missing_product_ids` from the in-progress data is load-bearing for the final `array_diff`, not just defensive. Job attrs round-trip through `wp_json_encode` / `json_decode( $raw, true )`, so the strict `in_array()` still matches after the restore (ints stay ints).

Second, this is the natural place to tell the merchant what happened, reusing the string the validate path already added (same translation entry, no new copy to review):

```suggestion
				if ( ! $product instanceof \WC_Product ) {
					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
					// out of the unprocessed set and the step can complete rather than re-queueing it
					// forever. Logged and recorded once per product (and not repeated on a resumed run).
					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
						$missing_product_ids[] = $product_id;

						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );

						// No product_id on the record: the post is gone, so a product-scoped record would
						// render an empty edit link.
						Records::set_record(
							array(
								'type'    => 'alert',
								'message' => sprintf(
									/* translators: %1$s - Product ID. */
									__( 'Product ID (%1$s) is excluded from sync because the product could not be found.', 'woocommerce-square' ),
									$product_id
								),
							)
						);
					}
					continue;
				}
```

includes/Sync/Manual_Synchronization.php
```diff
@@ -1130,6 +1164,18 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
 			}
 
+			// If nothing could be staged (e.g. every product in this batch was deleted mid-sync),
+			// there is nothing to send to Square. Clear progress and return early so the missing IDs
+			// are excluded from unprocessed and the step can complete instead of looping on an empty
+			// upsert request.
+			if ( empty( $batches ) ) {
+				$this->set_attr( 'in_progress_upsert_catalog_objects', null );
+
+				$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
```
faisal-alvi: `$batches` is also empty when the first product on its own exceeds `get_max_objects_total()` (the `else { break; }` above). In that case this returns the full ID set as `unprocessed` and the caller re-queues the same batch indefinitely - the same stall as before the patch, just without the empty API call, so not a regression. Since this early return is new though, naming the reason in the log keeps an oversized-product stall from looking like the deleted-product case:

```suggestion
			if ( empty( $batches ) ) {
				if ( empty( $missing_product_ids ) ) {
					// Nothing staged and nothing missing means the first product alone exceeds the
					// object limit, so the caller re-queues this same batch. Log it so this state is
					// not mistaken for the deleted-product case.
					wc_square()->log(
						'Upsert batch is empty and no products are missing - product #' . reset( $product_ids ) .
						' exceeds the max objects per batch (' . $this->get_max_objects_total() . ').'
					);
				}

				$this->set_attr( 'in_progress_upsert_catalog_objects', null );

				$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
				$result['missing']     = $missing_product_ids;
```

Whether that oversized product should be dropped or the job failed outright feels like its own ticket rather than this PR.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1130,6 +1164,18 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
+
+				return $result;
+			}
+
 			try {
```
faisal-alvi: `missing_product_ids` is currently persisted only after the request returns (line 1216). If the batch throws (rate limit, transport), the IDs collected in that pass are lost and get re-detected and re-logged on the next run, so "logged once per product" holds only for the non-throwing path. Writing the dedupe list before the request fixes that without changing resume semantics - `unprocessed_upsert_response` stays null, so a resumed run still re-runs the staging loop:

```suggestion
			// Persist the missing IDs before the request so a throw (rate limit, transport) does not
			// lose them and re-log the same products on the next run.
			$in_progress['missing_product_ids'] = $missing_product_ids;
			$this->set_attr( 'in_progress_upsert_catalog_objects', $in_progress );

			try {
```

includes/Sync/Manual_Synchronization.php
faisal-alvi: Trailing comma after the last `sprintf()` argument, which the sibling `sprintf()` right below does not have. Fine on PHP 7.4+, purely consistency.

```suggestion
								$matched_product_id
```

includes/Sync/Manual_Synchronization.php
faisal-alvi: This is the natural place to tell the merchant what happened - without it the only merchant-visible output for a mid-sync deletion is the generic `Product #<a href="">11068</a> could not be updated.` with a dead link (see the review summary). Reusing the string the validate path already added keeps it to one translation entry:

```suggestion
				if ( ! $product instanceof \WC_Product ) {
					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
					// out of the unprocessed set and the step can complete rather than re-queueing it
					// forever. Logged and recorded once per product (and not repeated on a resumed run).
					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
						$missing_product_ids[] = $product_id;

						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );

						// No product_id on the record: the post is gone, so a product-scoped record would
						// render an empty edit link.
						Records::set_record(
							array(
								'type'    => 'alert',
								'message' => sprintf(
									/* translators: %1$s - Product ID. */
									__( 'Product ID (%1$s) is excluded from sync because the product could not be found.', 'woocommerce-square' ),
									$product_id
								),
							)
						);
					}
					continue;
				}
```

includes/Sync/Manual_Synchronization.php
faisal-alvi: Second half of the same change: returning the missing IDs lets `upsert_new_products()` exclude them from the "could not be updated" loop, so a deleted product is not reported as an update failure. Caller-side diff is in the review summary, since those lines are not in this diff.

```suggestion
		$result['processed'] = $staged_product_ids;
		// Exclude missing products (deleted mid-sync) from unprocessed so callers do not re-queue
		// them indefinitely, which would keep the sync step from ever completing. They are also
		// returned so callers can tell "deleted" apart from "failed to update".
		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
		$result['missing']     = $missing_product_ids;
```

includes/Sync/Manual_Synchronization.php
faisal-alvi: Verified both scenarios live (Square sandbox, Woo as SOR): without the patch this line crashed with `get_meta()` on false and left the job at `status: processing`; with it the survivor upserts and the job completes.

One gap though: the skip is only visible in the debug log. The merchant-facing output for a mid-sync deletion ends up being the generic `Product #<a href="">11068</a> could not be updated.` from the end of `upsert_new_products()` - empty href, because `get_edit_post_link()` returns null once the post is gone, and no mention of the deletion. Worth recording the same alert the validate path added, so it reuses one translation entry, wdyt?:

```suggestion
				if ( ! $product instanceof \WC_Product ) {
					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
					// out of the unprocessed set and the step can complete rather than re-queueing it
					// forever. Logged and recorded once per product (and not repeated on a resumed run).
					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
						$missing_product_ids[] = $product_id;

						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );

						// No product_id on the record: the post is gone, so a product-scoped record would
						// render an empty edit link.
						Records::set_record(
							array(
								'type'    => 'alert',
								'message' => sprintf(
									/* translators: %1$s - Product ID. */
									__( 'Product ID (%1$s) is excluded from sync because the product could not be found.', 'woocommerce-square' ),
									$product_id
								),
							)
						);
					}
					continue;
				}
```

faisal-alvi, line 1169: Verified both scenarios live (Square sandbox, Woo as SOR): without the patch this line crashed with `get_meta()` on false and left the job at `status: processing`; with it the survivor upserts and the job completes.

One gap though: the skip is only visible in the debug log. The merchant-facing output for a mid-sync deletion ends up being the generic `Product #<a href="">11068</a> could not be updated.` from the end of `upsert_new_products()` - empty href, because `get_edit_post_link()` returns null once the post is gone, and no mention of the deletion. Worth recording the same alert the validate path added, so it reuses one translation entry:

```suggestion
				if ( ! $product instanceof \WC_Product ) {
					// Product is gone (e.g. permanently deleted mid-sync). Mark it missing so it drops
					// out of the unprocessed set and the step can complete rather than re-queueing it
					// forever. Logged and recorded once per product (and not repeated on a resumed run).
					if ( ! in_array( $product_id, $missing_product_ids, true ) ) {
						$missing_product_ids[] = $product_id;

						wc_square()->log( 'Skipping product #' . $product_id . ' during upsert - it no longer exists (may have been deleted).' );

						// No product_id on the record: the post is gone, so a product-scoped record would
						// render an empty edit link.
						Records::set_record(
							array(
								'type'    => 'alert',
								'message' => sprintf(
									/* translators: %1$s - Product ID. */
									__( 'Product ID (%1$s) is excluded from sync because the product could not be found.', 'woocommerce-square' ),
									$product_id
								),
							)
						);
					}
					continue;
				}
```

includes/Sync/Manual_Synchronization.php
faisal-alvi, line 1378: Other half of the same fix: return the missing IDs so `upsert_new_products()` can keep them out of the "could not be updated" loop.

```suggestion
		$result['processed'] = $staged_product_ids;
		// Exclude missing products (deleted mid-sync) from unprocessed so callers do not re-queue
		// them indefinitely, which would keep the sync step from ever completing. They are also
		// returned so callers can tell "deleted" apart from "failed to update".
		$result['unprocessed'] = array_diff( $product_ids, $staged_product_ids, $missing_product_ids );
		$result['missing']     = $missing_product_ids;
```
Please make sure to update other part of the codebase as needed after this change.

faisal-alvi (changes requested):
Thanks for the work @iamdharmesh 

Added a couple of feedback points below.

includes/Sync/Manual_Synchronization.php
iamdharmesh: Thanks @faisal. This should be good now.

includes/Sync/Manual_Synchronization.php
iamdharmesh: We are good here now. (Not using the missing in result but using the job attr instead)

faisal-alvi (approved):

qasumitbagthariya:
### QA Update  ✅ 
-------------

I have verified this PR in the `fix/delete-sync` branch, which has been fixed and is functioning as intended.

- Verified both product deletion scenarios during Square sync. Deleted products are now skipped correctly without causing fatal errors or a stuck sync, remaining products sync successfully, and the appropriate alert/log entry is generated.

All test scenarios passed successfully. ✅

image


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/delete-sync

</Details>


🔁 Next Step  -  Ready for UAT

Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

PR #540, opened 2026-08-20
Skip products that Square rejects so one product cannot fail the whole sync

faisal-alvi:
### All Submissions:

---

Square rejects a whole `batch-upsert` request when any single object in it is invalid, and the sync treated that rejection as a job level failure. One unsyncable product therefore stopped the entire sync, and because the same product failed the same way on every retry, the sync stayed stuck until the merchant found and fixed it themselves. The same shape applied to a failing category chunk and to an inventory batch that referenced a Square object which no longer exists, where every other change in the batch was lost with it.

Reporting was the other half of the problem. The merchant was told only that a product ID "could not be updated in Square", even though Square had returned a precise reason that the plugin discarded.

This PR makes a failure that belongs to one object skip that object and let the rest of the sync finish, and it reports what happened. A failure is only ever isolated when Square's own error code says the object's data is invalid. Anything else, including server errors, timeouts and permission problems, still fails the job loudly, because in those cases the write may already have been applied and retrying the objects one at a time would create duplicate catalog items in Square. Skipping is opted into, never assumed.

Where an error code alone cannot prove whose fault it is, a second check decides. Inventory failures are attributed by testing the batch's own catalog object IDs against the error text, so a stale mapping is dropped while a misconfigured location, which Square reports with the same code, fails loudly instead of silently discarding up to a hundred inventory updates. Similarly, an item option collision raised while building a payload asks the job to restart with refreshed option data, so it is re-thrown rather than blamed on a product that was never broken.

Skipped products are now reported with the product name, SKU, a link to the product, and Square's own reason, and every skip is also logged. The job finishes with a "completed with errors" outcome and a summary of how many products were skipped.

Deviation from an acceptance criterion, called out deliberately: SQUARE-31 asks that when a batch fails and the problem item cannot be identified, the sync log the error and continue to the next batch. That holds for the catalog paths. For inventory it does not, and intentionally so. Continuing there means discarding real stock updates for products that were never at fault, and the case that triggers it in practice is an account level location misconfiguration that would then stay hidden. That path fails loudly instead. Happy to change it if the tradeoff should go the other way.

Closes https://linear.app/a8c/issue/SQUARE-143/feature-request-square-if-product-has-an-issue-skip-the-product
Closes https://linear.app/a8c/issue/SQUARE-31/improve-error-handling-in-the-sync-process

### Notes for reviewers

1. **No workflow runs the unit suite, so nothing here is gated by CI.** Correcting an earlier version of this description which implied otherwise: `.github/workflows/` runs phpcs, php-compat, eslint, e2e, plugin-check and qit, and none of them run `composer test-unit`. The tests this PR adds are tracked, reviewable and runnable by hand, but they will not fail a check. Please run them locally if you want that signal (see note 4 for the one line needed to make the suite boot).
2. What the tests do and do not cover, stated plainly. Covered by the added PHPUnit tests: the error classifier, which alone decides whether one product is skipped or the whole job fails, and the inventory attribution check that separates a stale catalog mapping from an account level location error. Not covered by them, and only exercised by a standalone harness kept outside the repo: the job restart check in the staging path, and the wiring that makes the staging path consult the classifier at all. Removing either of those last two leaves the unit suite green, so anyone changing those blocks should verify by hand.
3. The isolation, inventory and end to end paths were verified against a live Square sandbox, driven through the real background job rather than by calling step methods with injected state. What was exercised: a mixed inventory batch where a stale mapping was dropped and the good updates confirmed by reading the counts back out of Square, a product Square rejects on its own data end to end, a location the token does not own, a variable product with two attributes and four variations, and a resumed cycle. A trunk versus branch comparison on the same product set was also run, which is where the intended behaviour change is visible.
4. `NOT_FOUND` is deliberately treated as an isolatable code even though Square also returns it for a location that does not belong to the account. The attribution check handles that second case separately. Removing `NOT_FOUND` from the list to fix a location error would stop stale catalog mappings from being isolated at all.
5. This PR does not touch the plugin bootstrap, so on a plain trunk checkout `composer test-unit` cannot start and the new tests will not run. `require_once` on the Composer autoloader returns `true` rather than the `ClassLoader` once something else has loaded it, which is always the case under PHPUnit, so the `addPsr4()` call after it fatals. That is pre-existing and reproduces on clean trunk, and it is already fixed on the branch for #529, so this PR relies on that fix rather than repeating it. Until #529 is in trunk, the unit job will fail to boot. To run the tests before then, apply the small local patch in this gist: https://gist.github.com/faisal-alvi/208838dce6238a14d481c6822cba8280
6. `composer test-unit` reports 1 skipped test. That is a pre-existing conditional in the abilities registrar test which only applies when the abilities loader is absent, and is unrelated to this work.
7. A skipped product still counts toward the sync progress figure, so a sync that skipped products can report 100 per cent complete. That is deliberate for now: the progress figure counts products the cycle consumed, and consuming a product is what stops the sync retrying it forever. The outcome is reported separately as a completed with errors notice naming each skipped product. Splitting progress from success touches the shared status calculation used by the Square to WooCommerce import as well, so it is left out of this change.
8. This branch shares a method with #529. The two compose, and a test merge of the sync file is clean, but #529 should merge first and this branch will be rebased onto it before review.

1. Connect the plugin to a Square sandbox and set the sync setting so WooCommerce is the system of record.
2. Create three simple products with SKUs and stock quantities, and let a sync complete so they exist in Square.
3. Give one of those products a name longer than 512 characters, which Square rejects, and leave the other two valid.
4. Run "Sync now" and let the job finish.
5. Confirm the job completes rather than failing, and that the sync records show one alert naming the offending product and SKU with Square's reason, plus a summary line stating that one product was skipped.
6. Confirm in Square that the other two products synced, and that no duplicate items were created for any of them.
7. For the inventory path, set a valid product's stored Square variation ID meta to a well formed but nonexistent ID, set fresh stock quantities on it and on a valid product, and run "Sync now" again.
8. Confirm the products with real mappings have their new quantities in Square, and that the product with the dead mapping is reported as skipped rather than the whole inventory batch being lost.
9. To check the loud failure path, temporarily set the configured location ID to one the token does not own and run a sync. Confirm the job fails and that no product is reported as skipped, since nothing here is the product's fault. Restore the location afterwards.

### Changelog entry

> Fix - Skip products that Square rejects during a sync instead of failing the whole sync, and report the name, SKU and reason for each skipped product.

### DIFF
--- a/includes/Sync/Manual_Synchronization.php
+++ b/includes/Sync/Manual_Synchronization.php
@@ -52,6 +52,52 @@ class Manual_Synchronization extends Stepped_Job {
 	/** @var int the limit for how many inventory changes can be made in a single request */
 	const BATCH_CHANGE_INVENTORY_LIMIT = 100;
 
+	/**
+	 * Square error codes that mean "this object's own data is invalid", and nothing else.
+	 *
+	 * Skipping an object has to be opted into by a known data error, never assumed: anything not
+	 * listed here fails the job instead, because a server error, timeout or permission problem may
+	 * have applied the write already and re-sending would duplicate it in Square. Codes come from
+	 * \Square\Models\ErrorCode. Filterable via wc_square_isolatable_error_codes.
+	 *
+	 * Before removing a code from this list because it produced a bad skip: some codes are
+	 * deliberately listed here and narrowed by a second gate at the call site, because Square
+	 * reuses one code for both an object level problem and a job level one. Removing the code
+	 * disables the object level handling and does not fix the job level case.
+	 *
+	 * - NOT_FOUND is both a stale catalog mapping and a location that does not belong to the
+	 *   account. push_inventory_changes_isolated() drops nothing unless Square named an object
+	 *   present in the chunk.
+	 * - INVALID_VALUE is both bad product data and an item option name collision, which
+	 *   API::create_options_and_values() turns into a whole job replay. The staging catch in
+	 *   upsert_catalog_objects() snapshots woocommerce_square_refresh_sync_cycle and rethrows
+	 *   when it changed.
+	 *
+	 * @var string[]
+	 */
+	const ISOLATABLE_ERROR_CODES = array(
+		'BAD_REQUEST',
+		'MISSING_REQUIRED_PARAMETER',
+		'INCORRECT_TYPE',
+		'INVALID_VALUE',
+		'INVALID_ENUM_VALUE',
+		'INVALID_ARRAY_VALUE',
+		'INVALID_TIME',
+		'VALUE_EMPTY',
+		'VALUE_TOO_LONG',
+		'VALUE_TOO_SHORT',
+		'VALUE_TOO_LOW',
+		'VALUE_TOO_HIGH',
+		'VALUE_REGEX_MISMATCH',
+		'ARRAY_EMPTY',
+		'ARRAY_LENGTH_TOO_LONG',
+		'ARRAY_LENGTH_TOO_SHORT',
+		'UNPROCESSABLE_ENTITY',
+		'REQUEST_ENTITY_TOO_LARGE', // the combined request was too big; one object per request is the fix.
+		'NOT_FOUND',
+		'CONFLICT',
+	);
+
 	/** @var int max SKU-based lookups per push_inventory step to avoid rate limits */
 	const MAX_SKU_LOOKUPS_PER_PUSH_STEP = 20;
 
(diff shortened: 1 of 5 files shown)


includes/Sync/Manual_Synchronization.php
```diff
@@ -1114,25 +1235,103 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
+				// Building the catalog payload can fail on a single product's local data (for
+				// example a product Catalog_Item rejects as invalid, or an object Woo_SOR refuses
+				// because it is not of type ITEM): record and skip that product instead of letting
+				// the exception end the whole sync (SQUARE-143 / SQUARE-31).
+				$refresh_requested_before = get_option( 'woocommerce_square_refresh_sync_cycle', false );
+
+				try {
+					if ( ! $product instanceof \WC_Product ) {
+						throw new \Exception( __( 'the product no longer exists locally', 'woocommerce-square' ) );
+					}
```
iamdharmesh: I think we should handle this gracefully, which I believe we already done in https://github.com/woocommerce/woocommerce-square/pull/538

includes/Sync/Manual_Synchronization.php
```diff
@@ -1288,30 +1520,332 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
+			} catch ( \Exception $product_exception ) {
+
+				$classification = $this->classify_sync_error( $product_exception );
+
+				// Not this product's data: either nothing will succeed (auth, permissions) or the
+				// write may already have been applied (server error, timeout). Recording a skip
+				// would be wrong in both cases, so fail the job. The job level retry re-sends this
+				// product under the same deterministic key, which Square deduplicates.
+				if ( 'fatal' === $classification ) {
+					throw $product_exception;
```
iamdharmesh: I think we should throw this only in case of "empty( $done_ids ) && empty( $failed_ids )" like we are doing for rate_limited to allow Mapping new Square item IDs to WooCommerce product IDs for the already processed products? what do you think?

includes/Sync/Manual_Synchronization.php
```diff
@@ -52,6 +52,55 @@ class Manual_Synchronization extends Stepped_Job {
 	/** @var int the limit for how many inventory changes can be made in a single request */
 	const BATCH_CHANGE_INVENTORY_LIMIT = 100;
 
+	/** @var int maximum per product fallback upserts attempted within one step cycle */
+	const MAX_ISOLATED_UPSERTS_PER_CYCLE = 20;
```
iamdharmesh: Can we reduce this to something like 5? This will only run during failure, so it's fine to have it smaller. The smaller will be safe against the timeout limit.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1114,25 +1235,103 @@ protected function upsert_catalog_objects( array $objects, $new_products = false
...
+				} catch ( \Exception $staging_exception ) {
+
+					// A failure that asked for the job to restart is never this product's fault.
+					// API::create_options_and_values() sets woocommerce_square_refresh_sync_cycle
+					// and clears the cached options data when Square rejects an item option, so
+					// that run() refetches the options and replays the cycle. Recording a skip here
+					// would strand a perfectly good product that the replay would have synced.
+					if ( get_option( 'woocommerce_square_refresh_sync_cycle', false ) !== $refresh_requested_before ) {
+						throw $staging_exception;
+					}
```
iamdharmesh, line 1302: Can we update the set "woocommerce_square_refresh_sync_cycle" only when option create fail due to it already exists and not due to something else? because this will fail the whole job.

iamdharmesh (changes requested):
Thanks @faisal-alvi. Great work here. I have done an initial review, not a full review yet. But I am assigning it back to you check on the initial feedback/notes to speedup the work here.

includes/Sync/Manual_Synchronization.php
faisal-alvi: Right, and #538's version is better than mine, so I removed mine rather than keeping both.
Yours tracks a persisted `upsert_missing_product_ids` set, so the alert fires exactly once per product per job even across a resume. Mine just threw and recorded inline.

I did not leave a gap in the meantime. The product now falls through to Catalog_Item's constructor, which rejects it with "Invalid product", and that is caught as a local build failure and recorded as a skip. So this PR standalone still handles a deleted product gracefully, and when #538 lands its explicit check runs first and does it better.

Whichever of the two merges second should drop its copy. Flagging it here so it does not get resolved twice.

includes/Sync/Manual_Synchronization.php
faisal-alvi: Would rather not, here. It is not what fails the job: `API::create_options_and_values()` sets the
flag on any failure regardless of my sentinel, so a 500 there fails identically before and
after this PR. And that method is also on the Square to WooCommerce import path, so narrowing
it changes recovery outside this ticket.

Happy to raise it separately. SQUARE-363 depends on the same replay mechanism, so it may be
cheaper to fix there.

includes/Sync/Manual_Synchronization.php
faisal-alvi: Agreed, and it removed a branch rather than adding one: fatal and rate limited now share one path that throws only when there is nothing to keep.

I think its safe because the per-product key has no timestamp, so re-sending next cycle reuses the same key and Square deduplicates. Auth still fails the job, one cycle later.

includes/Sync/Manual_Synchronization.php
```diff
@@ -1581,6 +2114,95 @@ protected function push_inventory() {
...
+	 * @param \Square\Models\InventoryChange[] $inventory_changes the changes to push
+	 */
+	protected function push_inventory_changes_isolated( array $inventory_changes ) {
+
+		foreach ( array_chunk( $inventory_changes, self::BATCH_CHANGE_INVENTORY_LIMIT ) as $chunk ) {
+
+			// Every round drops at least one change, so this terminates.
+			$attempts_left = count( $chunk ) + 1;
+
+			while ( ! empty( $chunk ) && $attempts_left > 0 ) {
```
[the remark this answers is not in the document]
faisal-alvi: Capped at 5 rounds. Your 101 assumes Square names one object per response; it names all of
them, which this PR already exploits by dropping every named ID per round.

Also fixed a hole nearby: exhausting the old counter silently abandoned the rest of the chunk.
It now throws. No resume state added, since those products are already off the queue for the
cycle so nothing loops.

iamdharmesh (approved):
Thanks @faisal-alvi. This looks good to me now. Approving with follow up note.

- For larger sync I noticed error reported in https://linear.app/a8c/issue/SQUARE-363/sync-fails-with-invalid-value-when-square-has-100-item-options and we are not skipping product for this error. So, better to resolve 363 or skip product instead of sync failure.
- Pushing new products seems super slow due to API rate limit errors can we figure out a ideal numbers to avoid this errors to do sync up to speed for the pushing new product? (In separate issue)
- I noticed set_record keep only last 50 records for larger sync better to increase it to keep all the alert available after sync (Separate issue)

Thanks for all the work here.

qasumitbagthariya:
### QA Update  ✅
---

I have verified this PR in the `fix/SQUARE-143-31` branch, which has been fixed and is functioning as intended.

- Verified the Square sync handling for invalid product data, dead variation mappings, and invalid location configuration. Invalid products/mappings are skipped with the appropriate alerts while valid products continue to sync successfully. Invalid location configuration correctly causes the sync to fail without incorrectly reporting products as skipped.


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/SQUARE-143-31

</Details>


🔁 Next Step  -  Ready for UAT

Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

PR #545, opened 2026-08-25
Reuse existing Square item options instead of failing on case or pagination

faisal-alvi:
### All Submissions:

---

Closes https://linear.app/a8c/issue/SQUARE-363/sync-fails-with-invalid-value-when-square-has-100-item-options.

`[INVALID_VALUE] An existing Item Option has name "X"` has two independent causes. The ticket describes the first one. Testing the customer catalogue against a production Square account surfaced the second, and either one on its own is enough to fail an entire sync.

**Cause 1, a paginated read kept only its last page.** `retrieve_options_data()` started its accumulator empty on every call and wrote the cache only once the cursor was exhausted, so on a store with more than 100 item options the pages before the last were discarded. Options that exist in Square then looked absent, the sync asked Square to create them, and Square refused. `fetch_options_data()` also dropped the page payload it was handed.

Pages now accumulate under their own transient key and the finished set is promoted to `wc_square_options_data` only when the cursor runs out, at which point the partial is deleted. The separate key is not incidental: the function returns the cached value early when it exists, so persisting partial pages under the finished key would hand back a half built cache and re-emit the same cursor indefinitely. The reset path that discards the finished cache after a failed create discards the partial with it, so an abandoned walk leaves nothing behind for a later write to land in.

**Cause 2, names were matched case sensitively.** Square treats item option names as case insensitive and rejects a create whose name differs from an existing one only by case, while the plugin compared with `===`. A store whose Square account holds `size` and whose products use a `Size` attribute could not sync at all, and no amount of merging pages would help, because the option was in the cache the whole time.

Option names, existing option values, and the values read back from a freshly upserted option are now compared with `strcasecmp`, and the value diff uses `array_udiff( ..., 'strcasecmp' )` so a reused option is not asked to create `large` alongside Square's own `Large`. Reuse is non destructive: `create_options_and_values()` with an ID takes the retrieve path and never calls `setName`, so nothing already in the merchant's account is renamed.

**A fatal on the same path.** Once names matched case insensitively, cached IDs pointing at options that no longer exist in Square began to resolve, and `create_options_and_values()` called `setValues()` on the resulting null item option, taking the sync down with `Uncaught Error: Call to a member function setValues() on null`. Being a fatal rather than an exception, no error isolation can catch it. An ID that Square cannot resolve to an `ITEM_OPTION` is now treated as no match and falls through to the create path. This looks like an incomplete fix for SQUARE-292, which guarded `getValues()` in three places but not this call.

**A second create on the variation path.** With option matching fixed, `update_catalog_variation()` was still able to ask Square for an option it had just been handed. Its option ID started empty and was only filled from the options cache, so when the cache did not yet know about the option the parent product had created, the call fell through to the create path carrying no ID, and Square rejected the duplicate name. The ID the parent product resolved is now carried into that call whatever the cache holds, so the create takes the retrieve path and reuses the option. This was reachable only while the finished cache was absent on a paginating account, and a reset cycle recovered from it, but it failed the same sync the rest of this branch is fixing.

**Case sensitivity, the cause seen on real merchant data. No special data volume needed.**

1. In the Square dashboard, create an item option named entirely in lower case, for example `size`, with a couple of values.
2. In WooCommerce, create a global or custom product attribute whose name differs only by case, for example `Size`, mark it used for variations, and give a variable product two variations using it.
3. Set System of Record to WooCommerce and run a manual sync of that product.
4. On trunk the sync fails with `[INVALID_VALUE] An existing Item Option has name "Size"`, and the failure repeats through three reset cycles before the job dies.
5. On this branch the sync completes, the product is created in Square, and the existing `size` option is reused. Confirm in the Square dashboard that no second option was created and that the original name is unchanged.

**Pagination, the cause in the ticket. Needs more than 100 item options in Square.**

1. On a Square account holding more than 100 `ITEM_OPTION` objects, so that `ListCatalog` returns a cursor, set System of Record to WooCommerce.
2. Create a variable product whose attribute name matches an item option Square returns on a page other than the last.
3. Run a manual sync of that product.
4. On trunk the sync fails with `[INVALID_VALUE] An existing Item Option has name "..."`. On this branch it completes and reuses the existing option.
5. With logging enabled, `fetch_options_data` reports a running option count that reaches the full number of options in the account rather than the size of the final page alone.

**Regression check, either scenario.**

1. Run a sync of a handful of variable products with mixed attribute casing and confirm no duplicate item options appear in Square.
2. Confirm existing option values are reused rather than duplicated in different cases.

### Verification already done

Measured against the On The Square customer catalogue on a 10up owned Square **production** account, System of Record WooCommerce:

- Before this branch, a 60 product scope (30 variable, 30 simple) failed at **0 processed** with the `Size` collision, and every product after the failure point never got its turn.
- On this branch the same scope **completes, 60 of 60 processed**, with 0 of 319 products changed in stock quantity, stock status or stock management, and 47 of 53 managed products carrying exact Square counts.
- The `setValues()` fatal reproduced during that work and no longer occurs.

That run predates the last two commits on this branch. Neither has been through a production sync: the partial cache clear is covered by the reasoning above, and the variation path change has no unit coverage at all, because `update_catalog_variation()` is static and needs WooCommerce loaded, which the unit harness does not provide. Both are lint and sniff clean, and the existing suite still passes with them applied.

### Unit tests

The pagination half cannot be exercised on our sandbox: the account holds six item options and Square's page size is fixed at 100, so no cursor is ever returned. It is covered by unit tests instead, against a scripted catalogue rather than the wire.

`tests/php/Unit/API/Scripted_API.php` replaces `list_catalog()`, `retrieve_catalog_object()` and `upsert_catalog_object()` with canned responses and skips the parent constructor, so no test can reach the network.

`tests/php/Unit/API/Retrieve_Options_Data_Test.php` covers the read:

- a two page and a three page walk keep every page, not just the last
- a later page wins for a repeated option ID
- a half read catalogue never lands in the finished cache
- a fresh read ignores a partial left behind by an abandoned one
- a complete cache still short circuits the walk, and a nameless cache is refetched and may still paginate
- the partial cache is given the lifetime of the finished one

`tests/php/Unit/API/Create_Options_And_Values_Test.php` covers the write back:

- existing values are matched without regard to case, and values differing only by case add nothing
- an unresolvable option ID, an ID resolving to another object type, and an ITEM_OPTION carrying no option data each fall through to the create path rather than reaching `setValues()` on null
- a created option is written back under its real ID, joins an in flight partial read when one exists, and is not published as a finished catalogue when neither cache exists
- the write back never reads the catalogue

There is no PHPUnit job in CI, so these run locally:

```
bash tests/php/bin/install-wp-tests.sh wordpress_test root '' localhost latest
composer test-unit
```

**Still not proven by execution:** a real Square account holding more than 100 item options. The behaviour is covered by the tests above and reasoned from the code, but no live cursor has been walked. The variation path change shares that gap, since the state it guards against only arises on a paginating account.

### Changelog entry

> Fix - Sync no longer fails when Square already holds an item option whose name differs only by case, or when a store has more than 100 item options.

### DIFF
--- a/includes/API.php
+++ b/includes/API.php
@@ -818,8 +818,25 @@ class API extends Base {
 
 			$id_mappings = $new_option->get_data()->getIdMappings();
 
-			if ( isset( $id_mappings[0] ) ) {
-				$option_id = $id_mappings[0]->getObjectId();
+			// Resolved on the create path only, and never positionally. The upsert reports a mapping
+			// for every object it created, so when an option that already existed merely gained a
+			// value, the first row is that new ITEM_OPTION_VAL and the option has no row at all.
+			// Index 0 therefore hands back a value ID, which travels on into the item's item_options
+			// and kills the following sync with "An existing Item Option has name X". On the reuse
+			// path $option_id is already the real ID Square gave us, so the mappings add nothing and
+			// must not be allowed to overwrite it.
+			if ( ! $option_id ) {
+				// The option went up under the temp ID set above, and that is the client ID Square
+				// echoes back alongside the real one. It is empty only when no attribute name was
+				// supplied, and there is nothing to correlate on then: guessing a row is the bug.
+				$client_option_id = $option->getId();
+
+				foreach ( (array) $id_mappings as $id_mapping ) {
+					if ( $client_option_id && $client_option_id === $id_mapping->getClientObjectId() ) {
+						$option_id = $id_mapping->getObjectId();
+						break;
+					}
+				}
 			}
 
 			$response = $this->retrieve_catalog_object( $option_id );
--- a/includes/API.php
+++ b/includes/API.php
@@ -692,6 +692,47 @@ class API extends Base {
 	}
 
 	/**
+	 * Records one item option in whichever options cache is authoritative right now.
+	 *
+	 * Callers here hold a single option they just read or created, not a walked catalogue. The
+	 * cache they must not touch is the finished one: building it from an unlooped read would pass
+	 * off the first page of a paginated catalogue as the whole of it, for a day. So a read still
+	 * in flight owns the data and the option joins its partial cache; otherwise the option is
+	 * folded into a finished cache that already exists. With neither, there is nothing to extend
+	 * and the next full read is left to build it.
+	 *
+	 * @since x.x.x
+	 *
+	 * @param string $option_id   Square item option ID.
+	 * @param array  $option_data Cache entry for the option.
+	 * @return void
+	 */
+	public function cache_option_data( $option_id, array $option_data ) {
+
+		if ( ! $option_id ) {
+			return;
+		}
+
+		$partial = get_transient( self::OPTIONS_DATA_PARTIAL_TRANSIENT );
+
+		if ( is_array( $partial ) ) {
+			$partial[ $option_id ] = $option_data;
+			set_transient( self::OPTIONS_DATA_PARTIAL_TRANSIENT, $partial, DAY_IN_SECONDS );
+
+			return;
+		}
+
+		$options_data = get_transient( 'wc_square_options_data' );
+
+		if ( ! is_array( $options_data ) ) {
+			return;
+		}
+
+		$options_data[ $option_id ] = $option_data;
+		set_transient( 'wc_square_options_data', $options_data, DAY_IN_SECONDS );
+	}
+
+	/**
 	 * Create options and values in Square.
 	 *
 	 * @since 4.9.0
@@ -777,15 +818,14 @@ class API extends Base {
 				$option_values[]    = $option_value->getItemOptionValueData()->getName();
 			}
 
-			$result       = $this->retrieve_options_data();
-			$options_data = isset( $result[1] ) ? $result[1] : array();
-
-			$options_data[ $option_id ] = array(
-				'name'      => $attribute_name,
-				'values'    => $option_values,
-				'value_ids' => array_combine( $option_value_ids, $option_values ),
+			$this->cache_option_data(
+				$option_id,
+				array(
+					'name'      => $attribute_name,
+					'values'    => $option_values,
+					'value_ids' => array_combine( $option_value_ids, $option_values ),
+				)
 			);
-			set_transient( 'wc_square_options_data', $options_data, DAY_IN_SECONDS );
 
 		} catch ( \Exception $e ) {
 			/**
--- a/includes/API.php
+++ b/includes/API.php
@@ -42,6 +42,17 @@ defined( 'ABSPATH' ) || exit;
 class API extends Base {
 
 
+	/**
+	 * Holds item option pages while the catalogue is still being walked.
+	 *
+	 * Kept apart from `wc_square_options_data`, which means "every option, fully read". Writing
+	 * partial pages under that name would make the next call return early with an incomplete cache.
+	 *
+	 * @since x.x.x
+	 * @var string
+	 */
+	const OPTIONS_DATA_PARTIAL_TRANSIENT = 'wc_square_options_data_partial';
+
 	/** catalog request type */
 	const REQUEST_TYPE_CATALOG = 'catalog';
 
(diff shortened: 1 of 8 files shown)


includes/Sync/Manual_Synchronization.php
```diff
@@ -2182,6 +2182,23 @@ protected function fetch_options_data() {
 		$result     = wc_square()->get_api()->retrieve_options_data( $cursor );
 		$new_cursor = isset( $result[2] ) ? $result[2] : null;
 
+		// Logged per page, not just at the end, so a read that stops short is visible here as a
+		// count that never reaches the number of options Square holds. Without it a truncated read
+		// only surfaces much later, as Square rejecting an option it already has.
+		if ( isset( $result[1] ) && is_array( $result[1] ) ) {
+			$option_count = count( $result[1] );
+
+			$this->set_attr( 'fetch_options_data_count', $option_count );
```
iamdharmesh: We are not using this anywhere, can we delete this?

includes/Sync/Product_Import.php
```diff
@@ -93,6 +93,23 @@ protected function fetch_options_data() {
 		$result     = wc_square()->get_api()->retrieve_options_data( $cursor );
 		$new_cursor = isset( $result[2] ) ? $result[2] : null;
 
+		// Logged per page, not just at the end, so a read that stops short is visible here as a
+		// count that never reaches the number of options Square holds. Without it a truncated read
+		// only surfaces much later, as Square rejecting an option it already has.
+		if ( isset( $result[1] ) && is_array( $result[1] ) ) {
+			$option_count = count( $result[1] );
+
+			$this->set_attr( 'fetch_options_data_count', $option_count );
```
iamdharmesh: same here.

includes/API.php
```diff
@@ -654,13 +677,61 @@ public function retrieve_options_data( $cursor = '', $refresh = false ) {
...
+		}
+
+		$partial = get_transient( self::OPTIONS_DATA_PARTIAL_TRANSIENT );
+
+		if ( is_array( $partial ) ) {
+			$partial[ $option_id ] = $option_data;
+			set_transient( self::OPTIONS_DATA_PARTIAL_TRANSIENT, $partial, DAY_IN_SECONDS );
+
+			return;
+		}
```
iamdharmesh: By the time this function called the full transient should already build and no partial options would be there? do we need this?

iamdharmesh (commented):
Thanks for the PR @faisal-alvi. I have added few comments based on only quick CR. will review in details later on.

includes/Sync/Manual_Synchronization.php
faisal-alvi: Good catch, it was write only. Confirmed nothing reads `fetch_options_data_count`, so it is removed here and in `Product_Import`.

includes/Sync/Product_Import.php
faisal-alvi: Removed in the same commit,https://github.com/woocommerce/woocommerce-square/pull/545/commits/6e5900083bd7ce75b77a8f88e7ad639452a0fbf5

includes/API.php
faisal-alvi: Yes, I checked, and you're correct; it is not needed. Steps run one at a time and `fetch_options_data` only finishes when the cursor is empty, so by the time any product step calls this the full transient is built and the partial is already deleted.

Removed in https://github.com/woocommerce/woocommerce-square/pull/545/commits/b31415ed3e70b9fabeeb2a13e1ec49e82fd75eea. `cache_option_data()` now just extends the finished transient if it exists, and does nothing if it does not.

iamdharmesh (approved):
Thanks @faisal-alvi. Changes looks good to me. 

Side note: The options are currently stored in a transient, which is not the correct storage mechanism for this use case, as it does not guarantee that the data will persist until the sync is completed. As a result, the sync could potentially fail if the transient is cleared. However, this is out of scope for this PR, so I’d suggest addressing it in a follow-up PR.

qasumitbagthariya:
QA Update: ✅
---
I have verified this PR in the `fix/SQUARE-363` branch, which has been fixed and is functioning as intended.

Tested against a real Square Production account (with SOR set to WooCommerce) so the item-option catalog already held 50+ existing options, giving the plugin's own pagination (limit=50) something to actually page through.

No regressions found in the sync steps this PR touches. The full production catalog sync (6,900+ products) was also run in parallel throughout testing without issue.


🖥 **Testing Environment** 

<Details>

- WordPress: 7.0.4
- Theme: Storefront 4.6.2
- WooCommerce - 11.0.1
- PHP: 8.0.30
- Web Server: Nginx 1.20.2
- Browser: Chrome
- OS: macOS 15.3.2
- Branch: fix/SQUARE-363

</Details>


🔁 Next Step  -  Ready for UAT

Testing Document status:

- [ ] Yes

qasumitbagthariya (approved):

PR #556, opened 2026-09-02
SQUARE-282: [Update UI] 5 - Migrate & Restyle: Synchronize Square Tab

faisal-alvi:
### All Submissions:

---

Square's product sync settings lived on the legacy Square settings screen, so a store manager had to leave the new settings hub to change how products and inventory flow between WooCommerce and Square.

This PR migrates them into the Synchronize Square tab of the modern Square settings hub (`WooCommerce > Settings > Payments > Square > Synchronize Square`) on WooCommerce 11.0+, restyled to the approved Figma design. Everything on the tab is persisted together when you click **Save**. The legacy settings screen is untouched and is still used whenever the modern settings path is inactive.

image

Closes https://linear.app/a8c/issue/SQUARE-282/update-ui-5-migrate-and-restyle-synchronize-square-tab

### Notes for QA/reviewers

**Two sections of the Figma design are intentionally not built here.**

* **Debug mode / Log synchronization messages.** Debug and logging is being refactored from the legacy seven option dropdown into a simpler model under [SQUARE-295](https://linear.app/a8c/issue/SQUARE-295), and is handled in that ticket. This matches the same decision taken on the Payments & Transactions tab.
* **Update and sync records / Go to Square records.** The sync records screen has no reachable home while the modern hub is active, so the button would have nowhere to point. Migrating that screen needs its own issue, which also covers the post import View Progress link.

1. On WooCommerce 11.0+ with the store connected to Square, go to `WooCommerce > Settings > Payments > Square > Synchronize Square` and confirm the three cards render and match the design.
2. Set Where the products are managed to Square and confirm Use Square product images and Hide missing products appear between the select and the interval. Set it to WooCommerce and confirm both hide. Set it to Disabled and confirm the inventory toggle, interval and the whole Manually import products section hide together.
3. Change one field at a time, click Save, reload, and confirm the value persisted and that no other setting on any Square tab changed. Confirm saving here does not reload the page.
4. Confirm the checkboxes round trip as `yes`/`no`, including that an unchecked box stays unchecked after a reload.
5. With unsaved changes on the page, confirm the import button is disabled and explains why. Save, then use Manually import products, confirm the modal opens, that Cancel does nothing, and that confirming starts a background import.
6. Force the legacy screen (filter `wc_square_is_modern_settings_active` to false), then confirm the legacy Square settings screen still renders the same sync settings and still saves.

### Changelog entry

> Add: Migrate the Square product sync settings into a Synchronize Square tab in the modern settings hub, covering inventory synchronization, order synchronization and Square discount codes.


includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -569,6 +580,195 @@ private function get_payment_methods_tab_groups(): array {
...
+							'value'       => wc_bool_to_string( wc_string_to_bool( $settings['enable_inventory_sync'] ?? 'no' ) ),
+						),
+						array(
+							'id'          => 'system_of_record',
+							'label'       => __( 'Where the products are managed', 'woocommerce-square' ),
+							'type'        => 'select',
+							'description' => sprintf(
+								/* translators: %1$s opening anchor tag, %2$s closing anchor tag */
+								__( 'Choose the official source of product and inventory data. %1$sLearn more about choosing a system of record ↗%2$s', 'woocommerce-square' ),
+								'<a href="https://woocommerce.com/document/woocommerce-square/#section-8" target="_blank" rel="noopener">',
```
iamdharmesh, line 637: This link not landing to the "system of record" doc.

src/settings/synchronize/import-products.js
```diff
@@ -0,0 +1,196 @@
...
+						) }
+					</h3>
+					<RawHTML>
+						{ sprintf(
+							/* translators: %1$s and %2$s are placeholders for the link to the documentation */
+							__(
+								'Doing so will update existing WooCommerce products with the latest information from Square. %1$sView Documentation%2$s.',
+								'woocommerce-square'
+							),
+							'<a href="https://woocommerce.com/document/woocommerce-square/#section-8" target="_blank" rel="noopener">',
```
iamdharmesh, line 163: Same here.

includes/Admin/Square_Modern_Settings_Page.php
```diff
@@ -569,6 +580,195 @@ private function get_payment_methods_tab_groups(): array {
...
+						),
+						array(
+							'id'          => 'hide_missing_products',
+							'label'       => __( 'Hide missing products', 'woocommerce-square' ),
+							'type'        => 'checkbox',
+							'description' => __( 'Products not found in Square will be hidden in the WooCommerce product catalog.', 'woocommerce-square' ),
+							'value'       => wc_bool_to_string( wc_string_to_bool( $settings['hide_missing_products'] ?? 'no' ) ),
+						),
+						array(
+							'id'          => 'sync_interval',
```
iamdharmesh, line 671: Changing **How often to sync** on the modern Synchronize tab persists the new value but never reschedules the recurring background sync, so the cadence silently does not change. The legacy screen reschedules via `Settings::maybe_change_sync_interval()`, called from the `woocommerce_settings_api_sanitized_fields_*` filter, which never fires on the REST save path. Could you please add this change sync interval on the new settings as well?

iamdharmesh (commented):
Thanks @faisal-alvi. This looks good overall. I have added few comments to check and could you please check and fix the valid copilot feedback as well? Thanks.

includes/Admin/Square_Modern_Settings_Page.php
faisal-alvi: Added in https://github.com/woocommerce/woocommerce-square/pull/556/commits/0c2b9714387131cadf05f2e3de33b42396fe06b3: the controller captures the old interval before the write and reschedules after `update_option()` and `init_settings()`.

